PR #38, opened 2019-02-14
add vpc_id to outputs

kilbergr:
I would like to add VPC ID to the outputs of the Atlantis terraform module.
Currently, when you spin up a VPC, AWS by default attaches a security group that has permissions that are broader than desirable. In order to remove this security group and replace it with one more appropriate, we'd like to have the VPC ID available to apply more changes after its creation.

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -18,6 +18,11 @@ output "task_role_arn" {
   value       = "${aws_iam_role.ecs_task_execution.arn}"
 }
 
+output "vpc_id" {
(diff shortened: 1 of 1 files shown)


antonbabenko:
Thanks @kilbergr !

[v1.12.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v1.12.0) has been just released.

kilbergr:
Thanks for doing so quickly @antonbabenko !

PR #40, opened 2019-03-04
replace /20 to /16 in README.md

KasteM34:
As an example /20 won't work as it covers until "10.20.15.255" (4096 possible IPs).


antonbabenko:
Good catch!

PR #60, opened 2019-06-12
fix task version

### LINKED ISSUE #59: Issues creating service on v2.0.0
https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/1195897966c4d0ab695d40c42865a4b0ebc19805/main.tf#L495

```
Error: Failed getting task definition ClientException: Unable to describe task definition.
        status code: 400, request id: 0cd48c37-8d35-11e9-9e00-5ff346cf8ee8 "atlantis"

  on .terraform/modules/atlantis/terraform-aws-modules-terraform-aws-atlantis-ca4ff06/main.tf line 491, in data "aws_ecs_task_definition" "atlantis":
 491: data "aws_ecs_task_definition" "atlantis" {
```

smiller171:
# Description

Fixes issues with creating the service caused by commenting out the `depends_on` in `data.aws_ecs_service.atlantis`. This also removes the ability to select the greater of the versions between what's in state and what data returns, but the version in state should be preferable to the version returned as data anyway. If there's a reason for this behavior that I'm not understanding please let me know.

Fixes #59

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -488,20 +488,10 @@ resource "aws_ecs_task_definition" "atlantis" {
   container_definitions = local.container_definitions
 }
 
-data "aws_ecs_task_definition" "atlantis" {
-  task_definition = var.name
-
-  //  Commented out to avoid always-changed status every time during "terraform apply"
-  //  depends_on      = [aws_ecs_task_definition.atlantis]
-}
-
 resource "aws_ecs_service" "atlantis" {
   name    = var.name
   cluster = module.ecs.this_ecs_cluster_id
-  task_definition = "${data.aws_ecs_task_definition.atlantis.family}:${max(
-    aws_ecs_task_definition.atlantis.revision,
-    data.aws_ecs_task_definition.atlantis.revision,
-  )}"
+  task_definition = "${aws_ecs_task_definition.atlantis.family}:${aws_ecs_task_definition.atlantis.revision}"
   desired_count                      = var.ecs_service_desired_count
   launch_type                        = "FARGATE"
   deployment_maximum_percent         = var.ecs_service_deployment_maximum_percent


antonbabenko:
Hi Scott!

In fact, it is necessary to have the `depends_on` to be able to release a new version of task definitions and not overwrite it with the one defined in this module.

It was like this there before, but I was not sure if it has anything to do with Terraform 0.12 and commented out it. Now I reverted it back in #62 and going to release in a minute.

antonbabenko:
[v2.1.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.1.0) has been released.

smiller171:
@antonbabenko Couldn't `prevent_destroy` be used instead?

antonbabenko:
It is not about destroying anything.

smiller171:
@antonbabenko I don't follow at all. I'm not using the trick you have here in my own code and it's successfully creating a new revision every time

antonbabenko:
If you delete "ECS Task Definitions" completely in AWS Console (or using aws-cli) you should see that code won't work without `depends_on`. It is needed when creating ECS Task Definition for the first time.

smiller171:
It works if you remove the dependency on the data source entirely like I did in this PR.

smiller171:
Gotcha. That's the bit I was missing. Is it really worth the pain of dealing with TF thinking it's always changed to support altering the task definition outside of terraform?

antonbabenko:
That's a really good question, Scott!

There are some solutions like this in nearly all big modules which are designed and implemented in a flexible way. Additionally, ECS support is often not-very-well supported in Terraform, so yes, we have to deal with solutions like this.

Terraform 0.12 makes our work on infrastructure modules easier if used properly. We just need to find common patterns for ~20 scenarios and apply them (we used to do this for all major releases of Terraform).

I think we are on the right path by supporting a wide variety of features and a lot more is coming...

PR #66, opened 2019-07-03
Add capability to attach additional security groups to ALB

amittag:
# Description

This change is to allow a list of security group IDs to be added to the load balancer. This will enable externally created security groups to be passed through to the ALB module.


robertomoutinho:
Can we get a code review here? I could use that feature.

antonbabenko:
[v2.2.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.2.0) has been just released with this feature.

Thanks @amittag!

PR #74, opened 2019-08-30
Allow specifying atlantis_url directly

heisian:
# Description

By default, this module will use the DNS name of the load balancer it creates.

This DNS name doesn't match the Route53 record created for atlantis, and thus will give an SSL error if you have created an ACM certificate for the Route53 record/zone.

This change allows one to override the `atlantis_url`.


bryantbiggs:
hello @heisian - would you mind merging master/rebasing to update your branch please, and change your PR title to something like `feat: allow specifying custom 'atlantis_url'`

heisian:
Hi @bryantbiggs - sorry, because the reply was almost a year later I no longer am working in the same space. Hopefully someone else can pick this up.

PR #75, opened 2019-10-18
Fix the atlantis_url variable name

berkil:
# Description

Please explain the changes you made here and link to any relevant issues.


berkil:
Opened it without realising it should be as its now

Issue #69, opened 2019-07-29
Provide option to make ALB internal


chadasapp:
What it says in the description. In our setup it would be ideal if the load balancer was internal, on the private subnets so it's not publicly routed. It seems like it'd be fairly straightforward to implement.

mlcooper:
I agree.  I've been working with this module for the first time this week and just about have it up and running in my existing AWS VPC.  What I did do is try to 'fool' it and gave it my private subnets for the ALB since I don't want this exposed externally.  So the value I gave it for `private_subnet_ids` is the same as `public_subnet_ids`.  I _think_ it should work, but I'm not quite ready to test yet.

mlcooper:
This worked...I had to fork this terraform-aws-atlantis repo and add
```
load_balancer_is_internal = true
```
to the `module "alb" {}` block.

geota:
any progress/appetite to merge a feature like this? i would like to lock down ALB to make it fully internal, happy to submit a PR

martinmosegaard:
Hi @geota . I am trying to get this working in a PR. Let me know what you think.

antonbabenko:
This has been fixed in #99. Please use [v2.9.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.9.0) which has been just released.

For some users, this release may require some ALB resource recreation described in [ALB upgrade guide 5.0](https://github.com/terraform-aws-modules/terraform-aws-alb/blob/master/UPGRADE-5.0.md).

kneufeld-pbp:
Was it by design to always use public subnets when internal is set to `true`?  Our assumption was the module would use the `private_subnet_ids` when we provided the `vpc_id`; this assumption will error and our work-around is to add the set the public_subnet_ids with our `private_subnet_ids`

antonbabenko:
@kneufeld-pbp This is just a bad name for the variable. Previously, there was no way to make ALB internal, so the public subnets were always used.

Issue #67, opened 2019-07-04
Task definition is broken with pre-created secrets


Vlaaaaaaad:
Hi,

Looks like something regarding the secrets is broken. We discovered this during our upgrade to `0.12` and the `v2.1.0` of this module.

The following config used to work:
```
module "atlantis" {
  source  = "terraform-aws-modules/atlantis/aws"
  version = "~> 2.1.0"

  name = "atlantis"

  # VPC
  cidr            = "10.20.0.0/16"
  azs             = ["us-east-1a", "us-east-1b", "us-east-1c"]
  private_subnets = ["10.20.1.0/24", "10.20.2.0/24", "10.20.3.0/24"]
  public_subnets  = ["10.20.101.0/24", "10.20.102.0/24", "10.20.103.0/24"]

  # DNS (without trailing dot)
  route53_zone_name = "example.com"
  # Atlantis
  atlantis_github_user       = "atlantis-bot"
  atlantis_repo_whitelist    = ["github.com/terraform-aws-modules/*"]
}

provider "aws" {
  region = "us-east-1"
}
```

Note that that's the same code from the example, without `atlantis_github_user_token` defined. We don't want to have that in code so we put it in SSM. 
This module is smart enough that it can take it from SSM if it's not in the Terrafrom config. That's what was happening in `v1.9.0`.

In the `v2.1.0` this leads to the following in the JSON task definition:
```
     "secrets": [
        {
          "valueFrom": "unknown_secret_name_value",
          "name": "unknown_secret_name_key"
        },
        {
          "valueFrom": "/atlantis/webhook/secret",
          "name": "unknown_secret_webhook_key"
        }
      ],
```
This is obviously wrong and leads to the Task definition failing cause it cannot find those SSM secrets.

I think in https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/master/main.tf#L17-L23 there's something bad. That breaks the definitions in https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/master/main.tf#L67-L81. Maybe instead of the `token` there should've been `user` as I define the user, but the `token` is always empty.

Am I missing something?

mbravorus:
+1

Complex unparenthesised conditionals added as part of Terraform 0.12 compatibility commit make completely no sense when trying to invoke the module in an existing infra. Secrets don't get populated, IAM policy which should allow the task to access SSM secrets doesn't get applied. 

[...]

PR #77, opened 2019-10-29
Allow more secure practices (no secret leaking, internal alb)

tjarjoura:
# Description

This patch does the following:

* Allows you to use pre-existing values from SSM instead of having to pass them in through the module (and thus storing your secrets in the state file!) (I believe this addresses https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/67)
* Allow ALB to be internal (https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/69)
* Adds support for the bitbucket_base_url argument (this allows users of Bitbucket Server to use this module as well).
* Adds a "git_provider" argument, which cleans up the logic on choosing which SSM Key/Value to use. As far as I could tell the "has_secrets" variable didn't really make sense, as Atlantis must always be used with one of the three providers and therefore must pull one of the three user tokens. There is no use case where it should be false.

Please let me know if anything needs to be changed and I'll try to respond in a timely manner. Would love to get this merged into mainline as I am currently dependent on these additional features.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -180,7 +180,7 @@ module "alb" {
   log_bucket_name     = var.alb_log_bucket_name
   log_location_prefix = var.alb_log_location_prefix
 
-  load_balancer_is_internal = true
+  load_balancer_is_internal = var.internal_alb
 
   https_listeners = [
     {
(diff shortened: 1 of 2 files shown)


jhebert-coursera:
Thanks for writing this! I hope it makes it in soon.

Szasza:
@tjarjoura thank you for the contribution. Would it be possible to resolve the conflicts?

Thank you again.

PR #85, opened 2019-11-21
Add option for additional security groups

mpescetto:
# Description

I need to attach a security group to the ECS service that isn't generated by this module. So I'm adding an optional variable for that.

Also adding an optional variable to override the task execution role name so that we can name it what we want.


mpescetto:
Closing due to inactivity.

PR #91, opened 2020-01-14
Allow specifying Atlantis command line

sftim:
# Description

Allow specifying `CMD` for the Atlantis container inside Fargate.

This is useful if you want to set [custom flags](https://www.runatlantis.io/docs/server-configuration.html#flags) for the Atlantis server process.


lkysow:
You can already set `custom_environment_variables`. Every flag can be set as an environment variable so that should work?

sftim:
That is a good point @lkysow

bryantbiggs:
hi @sftim - this functionality is available now with v2.21.0 per #144

Issue #72, opened 2019-08-22
task_definition version always known after apply


TarekAS:
Hello,

While using this module, I get the following message every time I do ```terraform apply```.

```
~ resource "aws_ecs_service" "atlantis" {
        cluster                            = "arn:aws:ecs:eu-west-1:[REDACTED]:cluster/atlantis"
        deployment_maximum_percent         = 200
        deployment_minimum_healthy_percent = 50
        desired_count                      = 1
        enable_ecs_managed_tags            = false
        health_check_grace_period_seconds  = 0
        iam_role                           = "aws-service-role"
        id                                 = "arn:aws:ecs:eu-west-1:[REDACTED]:service/atlantis"
        launch_type                        = "FARGATE"
        name                               = "atlantis"
        platform_version                   = "LATEST"
        propagate_tags                     = "NONE"
        scheduling_strategy                = "REPLICA"
        tags                               = {}
      ~ task_definition                    = "atlantis:2" -> (known after apply)
```

Is there any way to avoid this?

The following is my module usage:

```
module "atlantis" {
  source  = "terraform-aws-modules/atlantis/aws"
  version = "~> 2.0"

  name               = "atlantis"
  vpc_id             = "${aws_vpc.myvpc.id}"
  private_subnet_ids = ["${aws_subnet.private-1a.id}", "${aws_subnet.private-1b.id}", "${aws_subnet.private-1c.id}"]
  public_subnet_ids  = ["${aws_subnet.public-1a.id}", "${aws_subnet.public-1b.id}", "${aws_subnet.public-1c.id}"]

  certificate_arn       = "redacted"
  create_route53_record = false

  atlantis_github_user       = "${local.atlantis_github_username}"
  atlantis_github_user_token = "${local.atlantis_github_access_token}"
  atlantis_repo_whitelist    = ["redacted"]

  ssm_kms_key_arn = "redacted"
}
```

jondunning:
Having the same issue

jseiser:
Same issue present in my deployments as well.

wreulicke:
I have same issue.

I seem that this issue is caused by this PR.
https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/62

wreulicke:
@antonbabenko 
What do you think?

tomwj:
https://github.com/tomwj/terraform-aws-atlantis/commit/22a4b644bcc14447ca6d209013e8055c88561d21

[...]

PR #124, opened 2020-05-13
Fix issue #72 task_definition known after apply

tomwj:
## Description
Using a data provider means terraform doesn't have the information required until runtime to determine if the `task_definition` needs to be updated. Subsequently an update is triggered on every `apply`. It has the information locally already, by removing the indirection of a data provider terraform can correctly reconcile if any updates are required prior to `apply` being run. Essentially it's making the dependency between the `aws_ecs_service` and `aws_ecs_task_definition` explicit rather than implied through the `data` provider and `depends_on`

https://www.terraform.io/docs/configuration/data-sources.html#data-resource-dependencies

## Motivation and Context
Currently atlantis shows an update everytime a plan or apply is run despite no configuration changes.
https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/72

## Breaking Changes
None that I'm aware of.

## How Has This Been Tested?
- Confirmed that updating the vars that are used in the `task_definition` trigger an update
- Manually create a new task_definition revision and confirm that terraform will role back to the 

AWS, plan/apply and manual changes

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -501,19 +501,10 @@ resource "aws_ecs_task_definition" "atlantis" {
   tags = local.tags
 }
 
-data "aws_ecs_task_definition" "atlantis" {
-  task_definition = var.name
-
-  depends_on = [aws_ecs_task_definition.atlantis]
-}
-
 resource "aws_ecs_service" "atlantis" {
   name    = var.name
   cluster = module.ecs.this_ecs_cluster_id
-  task_definition = "${data.aws_ecs_task_definition.atlantis.family}:${max(
-    aws_ecs_task_definition.atlantis.revision,
-    data.aws_ecs_task_definition.atlantis.revision,
-  )}"
+  task_definition = "${aws_ecs_task_definition.atlantis.family}:${aws_ecs_task_definition.atlantis.revision}"
   desired_count                      = var.ecs_service_desired_count
   launch_type                        = "FARGATE"
   deployment_maximum_percent         = var.ecs_service_deployment_maximum_percent


antonbabenko:
This PR is not going to work for the situation when a user deploys Atlantis once and then uses something else to update ECS Task Definition (eg, deploy new version of an image).

I remember that I and some other people I know used [ecs-deploy](https://github.com/silinternational/ecs-deploy) or similar tools to do actual deployments. This way we were sure that all infrastructure was managed as code and the image was updating during CI/CD process. I may not remember all details now, but in general, this was the process. I don't remember what is the right approach to avoid what you are experiencing.

/cc @maartenvanderhoef @osterman Do you remember what is the right solution to achieve a clean plan with `task_definition` every time?

maartenvanderhoef:
Hi, what if you only drop the `depends_on` from the `aws_ecs_task_definition` datasource, IIRC `depends_on` inside datasources causes weird behaviour like this.

domcleal:
> Hi, what if you only drop the `depends_on` from the `aws_ecs_task_definition` datasource, IIRC `depends_on` inside datasources causes weird behaviour like this.

I think `depends_on` is needed for the first `apply` of this module to work, as the data source will fail if the task definition didn't already exist. Unfortunately data sources that depend on resources will always cause this behaviour of showing up in plans.

> This PR is not going to work for the situation when a user deploys Atlantis once and then uses something else to update ECS Task Definition (eg, deploy new version of an image).

I've opened #163 which keeps this working with a new input variable, is this a reasonable compromise do you think?

PR #125, opened 2020-05-15
Add a configuration for the Github Hostname

thyming:
## Description
This adds a variable to the module to allow targeting an on-prem installation of Github Enterprise. 

## Motivation and Context
Using this with enterprise requires manually injecting an environment variable.


antonbabenko:
Hi @thyming !

I am not sure we need to add this to the module because there are a lot of optional parameters like this.

You can always pass extra variables this like this:

```terraform
    {
      name  = "ATLANTIS_GH_HOSTNAME"
      value = "https://mygithub.com"
    }
]
```

Are you sure that the default value should be

thyming:
Fair enough. Just seemed inconsistent that there was one for gitlab and not github.

PR #135, opened 2020-06-20
feat: Refactor to use SSM ARNs directly from resources

DrFaust92:
## Description

remove the `aws_ssm_path` variable to get full proper ARN from SSM resources instead of building it manually

## Breaking Changes

NONE

## How Has This Been Tested?

Run with the sample tfvars, added the changes and re-run to get no diffs (aside from regular ECS diff)

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -382,11 +382,11 @@ data "aws_iam_policy_document" "ecs_task_access_secrets" {
   statement {
     effect = "Allow"
 
-    resources = coalescelist(
+    resources = flatten(list(
       aws_ssm_parameter.webhook.*.arn,
       aws_ssm_parameter.atlantis_github_user_token.*.arn,
       aws_ssm_parameter.atlantis_gitlab_user_token.*.arn,
-    aws_ssm_parameter.atlantis_bitbucket_user_token.*.arn)
+    aws_ssm_parameter.atlantis_bitbucket_user_token.*.arn))
 
     actions = [
       "ssm:GetParameters",
--- a/main.tf
+++ b/main.tf
@@ -98,8 +98,6 @@ locals {
 
 data "aws_region" "current" {}
 
-data "aws_caller_identity" "current" {}
-
 data "aws_route53_zone" "this" {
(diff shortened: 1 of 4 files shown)


main.tf
```diff
@@ -384,12 +382,11 @@ data "aws_iam_policy_document" "ecs_task_access_secrets" {
   statement {
     effect = "Allow"
 
-    resources = [
-      "arn:${var.aws_ssm_path}:ssm:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:parameter${var.webhook_ssm_parameter_name}",
-      "arn:${var.aws_ssm_path}:ssm:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:parameter${var.atlantis_github_user_token_ssm_parameter_name}",
-      "arn:${var.aws_ssm_path}:ssm:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:parameter${var.atlantis_gitlab_user_token_ssm_parameter_name}",
-      "arn:${var.aws_ssm_path}:ssm:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:parameter${var.atlantis_bitbucket_user_token_ssm_parameter_name}",
-    ]
+    resources = coalescelist(
```
antonbabenko, line 385: This will find the first non-empty list.

Maybe unrelated question... Is it possible to make this module work for multiple types of VCS at the same time with little or no changes?

DrFaust92: fixed this with bt replacing `coalescelist` with `flatten(list())` this works fine for create. but it didnt seem to work right with update. it only saw diff for update the second run

DrFaust92: to the other question, i don't think so... there are things like `secret_name_key` that are specific to the input and will default to one of the first parameters that are listed

antonbabenko:
Thanks, @DrFaust92 !

[v2.18.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.18.0) has been just released.

PR #136, opened 2020-06-23
feat: Add tags to aws_ecs_service resource

joakimhew:
## Description

Includes the `local.tags` in the `aws_ecs_service` resource.

## Motivation and Context

For consistency, tags should be used across all the compatible resources created by the module. `aws_ecs_service` seems to be the only resource missing tags.

## Breaking Changes

`no`

## How Has This Been Tested?

`applied locally`

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -550,6 +550,8 @@ resource "aws_ecs_service" "atlantis" {
     container_port   = var.atlantis_port
     target_group_arn = element(module.alb.target_group_arns, 0)
   }
+    
+  tags = local.tags
 }
 
 ###################


antonbabenko:
Thanks, @joakimhew !

[v2.19.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.19.0) has been just released.

nitrocode:
The breaking change is that this drops support for short arns so the global setting has to be enabled so tasks and clusters can be tagged, then the current setup has to be shut down, and then recreated

Issue #141, opened 2020-07-23
Thoughts on adding a tag to the public ingress security groups?


nitrocode:
We audit our security groups and for the ones that are open to the whole world, we add `outside_access=justified` tag. This prevents our tag from showing up in our daily audits.

Since this tf module doesn't support an additional tag on just the security group without affecting all resources, we've been applying the terraform and then running the following.

```
aws ec2 create-tags --resources \
  sg-atlantis_sg \
  sg-alb_https_sg \
  sg-alb_http_sg \
  --tags Key=outside_access,Value=justified
```

@antonbabenko thoughts on creating a new input to add additional tags to merge with the following module's `tags` argument?

* `module.atlantis_sg`
* `module.alb_https_sg`
* `module.alb_http_sg`

Something like `var.tags_sgs_additional`. I put in a PR for this and linked it to this ticket.

PR #128, opened 2020-05-26
feat: add ability to specify existing IAM role for ECS task execution

druchoo:
## Description

Add ability to specify existing IAM role for ECS task execution.

## Motivation and Context

Allow module to be used in scenarios where IAM roles cannot be created or already exist.

## Breaking Changes

No.

## How Has This Been Tested?

```HCL
module "atlantis" {
  ...

  ecs_task_execution_iam_role_arn = "arn:aws:iam::xxxxxxxxxxxx:role/xxxxxxxxxx"
}
```

PR #142, opened 2020-07-23
feat: Additional tags to wide open security groups

nitrocode:
## Description
Adds a new variable `var.tags_sgs_additional` to merge with the tags supplied on the security group resources that are open to the world. This is to add tags to those wide open security groups to pass our audits.

## Motivation and Context
For auditing purposes

Closes https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/141

## Breaking Changes
None

## How Has This Been Tested?
Locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,8 @@ allow_github_webhooks        = true
 | acm\_certificate\_domain\_name | Route53 domain name to use for ACM certificate. Route53 zone for this domain should be created in advance. Specify if it is different from value in `route53_zone_name` | `string` | `""` | no |
 | alb\_authenticate\_cognito | Map of AWS Cognito authentication parameters to protect ALB (eg, using SAML). See https://www.terraform.io/docs/providers/aws/r/lb_listener.html#authenticate-cognito-action | `any` | `{}` | no |
 | alb\_authenticate\_oidc | Map of Authenticate OIDC parameters to protect ALB (eg, using Auth0). See https://www.terraform.io/docs/providers/aws/r/lb_listener.html#authenticate-oidc-action | `any` | `{}` | no |
+| alb\_http\_security\_group\_tags | Additional tags to put on the http security group | `map(string)` | `{}` | no |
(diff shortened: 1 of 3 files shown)


bryantbiggs:
thanks @nitrocode - could you split this up to 3 tags since there are 3 security groups and name them:
- `alb_https_security_group_tags`
- `alb_http_security_group_tags`
- `atlantis_security_group_tags`

and the variables can be defined similar to what has been done on the `vpc` module [here](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/75a18883c55a709ac8b565557f71ca364701c732/variables.tf#L1722)

nitrocode:
Hi @bryantbiggs .

>- alb_https_security_group_tags
>- alb_http_security_group_tags
>- atlantis_security_group_tags

The way these are named, are you advocating to not do a `merge` on those resources' tags' regarding the `tags` variable and instead supplant `var.tags` for those resources ?

>and the variables can be defined similar to what has been done on the `vpc` module here

If I used the example you linked to, the variable in this PR would be `var.sgs_tags` and would apply to all of the open-to-the-world security groups. Is that correct?

bryantbiggs:
no no, I think merging is good. its common across the modules here, there is a default tags group and then individual tags per "resource". so if you look at like the VPC module as an example, you'll see they merge the default tags as well as tags for the specific resource like:

`var.tags,
    var.dhcp_options_tags,

nitrocode:
@bryantbiggs makes sense to me. I updated the changes to reflect your suggestions.

From previous comments by @antonbabenko, it looks like these PRs won't get any love until August https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/128#issuecomment-660943650.

bryantbiggs:
@nitrocode would you mind updating your branch with master - @antonbabenko 👍

nitrocode:
All set!

antonbabenko:
Thanks, @nitrocode !

[v2.22.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.22.0) has been just released.

PR #149, opened 2020-08-17
feat: add variable to set SSL listener policy

bryantbiggs:
## Description
- add variable to allow changing SSL listener policy on HTTPS listener; currently set to current modules default so no changes to current module users

## Motivation and Context
- to be able to use a stronger cipher suite on the ALB that is created

## Breaking Changes
none

## How Has This Been Tested?
- complete example was updated and validated deployment

### DIFF
--- a/examples/github-complete/main.tf
+++ b/examples/github-complete/main.tf
@@ -69,7 +69,7 @@ module "atlantis" {
   atlantis_allowed_repo_names = var.allowed_repo_names
 
   # ALB access
-  alb_ingress_cidr_blocks         = [var.personal_ip]
+  alb_ingress_cidr_blocks         = [var.alb_ingress_cidr_blocks]
   alb_logging_enabled             = true
   alb_log_bucket_name             = module.atlantis_access_log_bucket.this_s3_bucket_id
   alb_log_location_prefix         = "atlantis-alb"
--- a/examples/github-complete/main.tf
+++ b/examples/github-complete/main.tf
@@ -69,10 +69,11 @@ module "atlantis" {
   atlantis_allowed_repo_names = var.allowed_repo_names
 
   # ALB access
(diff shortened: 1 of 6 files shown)


examples/github-complete/main.tf
```diff
@@ -69,10 +69,11 @@ module "atlantis" {
   atlantis_allowed_repo_names = var.allowed_repo_names
 
   # ALB access
-  alb_ingress_cidr_blocks = [var.personal_ip]
-  alb_logging_enabled     = true
-  alb_log_bucket_name     = module.atlantis_access_log_bucket.this_s3_bucket_id
-  alb_log_location_prefix = "atlantis-alb"
+  alb_ingress_cidr_blocks         = [var.personal_ip]
```
antonbabenko, line 72: `var.personal_ip` is a bit _small_. It can be the whole set of CIDR blocks. Let's rename it to match, something like - `var.alb_ingress_cidr_blocks`

main.tf
```diff
@@ -198,6 +198,7 @@ module "alb" {
     prefix  = var.alb_log_location_prefix
   }
 
+  listener_ssl_policy_default = var.alb_listener_ssl_policy_default
```
antonbabenko, line 201: Fix formatting, please.

bryantbiggs: which formatting? this passes `tf fmt` and the pre-commit hook

examples/github-complete/main.tf
bryantbiggs: updated

antonbabenko:
Thank you!

[v2.23.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.23.0) has been just released.

main.tf
antonbabenko: my bad :)

Issue #152, opened 2020-08-23
Cannot tag ecs service unless long arns are enabled in the aws account


nitrocode:
It would be nice to support older AWS accounts that cannot easily migrate to the long arn format. 

In order to do this, we could add a new variable such as `var.use_old_arn_format` or `var.disable_tagging_ecs_service` or similar.

```
variable "use_old_arn_format" {
  type        = bool
  description = "A flag to enable/disable tagging the ecs resources that require the new longer arn format"
  default     = false
}
```

and a tag toggle to disable tagging of the ecs service.

```
  tags = var.use_old_arn_format ? null : local.tags
```

PR #153, opened 2020-08-23
feat: Only tag ecs service if longer arns are enabled in the aws account

nitrocode:
## Description
Closes https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/152

## Motivation and Context
Older AWS accounts that have not enabled the longer arn format will not be able to tag the ecs service. This will allow disabling tagging of the ecs service using a boolean.

## Breaking Changes
None

## How Has This Been Tested?
Locally


SteveKekacs (approved):
Lgtm -- also need this.

nitrocode:
@antonbabenko ping

variables.tf
```diff
@@ -523,3 +523,9 @@ variable "security_group_ids" {
   type        = list(string)
   default     = []
 }
+
+variable "use_old_arn_format" {
+  type        = bool
+  description = "A flag to enable/disable tagging the ecs resources that require the new longer arn format"
```
flaudisio, line 529:
```suggestion
  description = "A flag to enable/disable tagging the ecs resources that require the new longer arn format"
  type        = bool
```

Minor suggestion, just to follow the code styling :smile:

DrFaust92:
FYI, just a nit. if tags are changed out of state this will also cause a problem when it tries to remove tags from an ecs resource.

nitrocode:
> FYI, just a nit. if tags are changed out of state this will also cause a problem when it tries to remove tags from an ecs resource.

True but the whole point of terraform is to do everything in code. Anything outside of code is a workaround.

main.tf
```diff
@@ -608,7 +608,7 @@ resource "aws_ecs_service" "atlantis" {
     }
   }
 
-  tags = local.tags
+  tags = var.use_old_arn_format ? null : local.tags
```
bryantbiggs, line 611: I don't have any way of testing since I don't have any accounts that do not use the newer format, but what if we do this instead (related to this [comment](https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/153#issuecomment-703212662))
```suggestion
  tags = var.use_old_arn_format ? {} : local.tags
```

nitrocode: It's always better to do `null` since that safely removes the argument from being set in the first place.

We've been using this method in the https://github.com/cloudposse/terraform-aws-ecs-alb-service-task module with positive results.

bryantbiggs: I don't believe it always removes, which is why I was suggesting an empty map which removes any ambiguity. I do agree that `null` works great as a default variable value, but this is something slightly different - a different edge case. either way, we can try this approach. one thing though - I think we should adjust the variable name to `use_ecs_old_arn_format` to make it a bit more clear

nitrocode: > I don't believe it always removes

I tested this with 0.12.x, 0.13.x, and 0.14.x and it seems to remove the argument completely from the resource.

> either way, we can try this approach.

Ok 😄 

> one thing though - I think we should adjust the variable name to use_ecs_old_arn_format to make it a bit more clear

Agreed, merge conflicts fixed and that change was committed.

bryantbiggs (approved):
@antonbabenko 👍🏼

DrFaust92:
@nitrocode can you rebase?

DrFaust92 (approved):
@antonbabenko LGTM

antonbabenko:
Thanks! [v2.41.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.41.0) has been just released.

PR #161, opened 2020-10-26
feat: enable configuring of atlantis server --checkout-strategy

Mitsuwa:
## Description

Enable configuring `--checkout-strategy` Implemented in this atlantis pr https://github.com/runatlantis/atlantis/pull/427/files

```
		description: "How to check out pull requests. Accepts either 'branch' (default) or 'merge'." +
			" If set to branch, Atlantis will check out the source branch of the pull request." +
			" If set to merge, Atlantis will check out the destination branch of the pull request (ex. master)" +
			" and then locally perform a git merge of the source branch." +
			" This effectively means Atlantis operates on the repo as it will look" +
			" after the pull request is merged.",
```

## Motivation and Context

Atm an atlantis instance deployed via this module cannot configure this setting.  Meaning that all deployments from atlantis from this module will take the default `branch` option for this flag on `atlantis server`.

According to the docs, that will only check out the branch we wish to apply.  This can cause problems if someone made a branch prior to an important master change, then atlantis applies removing that last change (if the user never rebased master)

We would like to be able to try to use this flag on our atlantis instance.

## Breaking Changes

There are no breaking changes here, the variable that is introduced `atlantis_checkout_strategy` defaults to `branch` which is what the flag already defaults to `branch`, you can see in their pr or on latest master https://github.com/runatlantis/atlantis/blob/master/cmd/server.go#L157

## How Has This Been Tested?

These create environment vars in the same methodology this repo currently uses and these env vars are passed via viper.AutomaticEnv, they are also checked for in their tests on that pr 

https://github.com/runatlantis/atlantis/pull/427/files#diff-3b1c14f936e65b2202c014583477ef8d2a67f69d9946a6b4e6829dee5486d0c9R595

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ allow_github_webhooks        = true
 | atlantis\_bitbucket\_user | Bitbucket username that is running the Atlantis command | `string` | `""` | no |
 | atlantis\_bitbucket\_user\_token | Bitbucket token of the user that is running the Atlantis command | `string` | `""` | no |
 | atlantis\_bitbucket\_user\_token\_ssm\_parameter\_name | Name of SSM parameter to keep atlantis\_bitbucket\_user\_token | `string` | `"/atlantis/bitbucket/user/token"` | no |
+| atlantis\_checkout\_strategy | Configure atlantis server --checkout-strategy 'merge' or 'branch' | `string` | `"branch"` | no
 | atlantis\_fqdn | FQDN of Atlantis to use. Set this only to override Route53 and ALB's DNS name. | `string` | `null` | no |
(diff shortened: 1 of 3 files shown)


Mitsuwa:
@antonbabenko if you can take a look when you get a chance please 🙏

bryantbiggs:
hi @Mitsuwa - because we expose a catch all `var.custom_environment_variables` I don't think this change is necessary https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/05b9dc79cec1e68fffd19c0914fcb525098ab9a9/main.tf#L485

you can simply add your desired additional environment variables to that variable and it will get passed in

Mitsuwa:
Oh sweet i mustve missed that, thanks!

PR #151, opened 2020-08-23
feat: Use atlantis bot instead of a github user

nitrocode:
## Description
See https://www.runatlantis.io/docs/access-credentials.html#generating-an-access-token

This is a new feature as of 0.14.0

This PR closes https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/147

## Motivation and Context
It's much better to set this up as a github app instead of a github user as you have an easier interface for granular controls, app can be installed across an org, and app settings can control which repos have access to it.

## Breaking Changes
None

## How Has This Been Tested?
1. Destroyed current atlantis
1. Recreated atlantis
1. Opened PR
1. Tested PR using a github app instead of a github user

PR #166, opened 2020-11-22
fix: secret policy not created with github bot

kieranbrown:
## Description

With the addition of https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/151, we can now specify an `atlantis_github_webhook_secret`, however, with only this parameter the policy to access the secret in SSM is never created.

The policy will only be created when `atlantis_github_user_token` is specified, which isn't necessary for a GitHub Bot.

## Motivation and Context
To run Atlantis as a Github App you only need to supply 3 variables, `ATLANTIS_GH_APP_ID`, `ATLANTIS_GH_APP_KEY_FILE` and `atlantis_github_webhook_secret`. With this setup, Atlantis should be able to access the webhook secret via SSM.

## Breaking Changes
n/a

## How Has This Been Tested?
Specifying an `atlantis_github_webhook_secret` will now attach the appropriate policy.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -424,7 +424,7 @@ data "aws_iam_policy_document" "ecs_task_access_secrets_with_kms" {
 }
 
 resource "aws_iam_role_policy" "ecs_task_access_secrets" {
-  count = var.atlantis_github_user_token != "" || var.atlantis_gitlab_user_token != "" || var.atlantis_bitbucket_user_token != "" ? 1 : 0
+  count = var.atlantis_github_webhook_secret != "" || var.atlantis_github_user_token != "" || var.atlantis_gitlab_user_token != "" || var.atlantis_bitbucket_user_token != "" ? 1 : 0
 
   name = "ECSTaskAccessSecretsPolicy"
 


DanielWright:
Apologies for posting to a presumably stale issue, but I'm fairly sure #151 doesn't add support for the GitHub App authentication, as there's no affordance for supplying the app ID or private key in the TF module. Maybe I'm missing something?

thundering-herd:
@DanielWright AFAIK this is correct. The workflow to install `atlantis` as a GitHub App and use it is following:

- run `atlantis` visit the endpoint `https://$ATLANTIS_HOST/github-app/setup`
- install `atlantis` into your GitHub tenant
- GitHub returns the App ID and a private key
- inject those value and private key into `atlantis` on the next start

So actually we are missing:

- `ATLANTIS_GH_APP_ID`
- `ATLANTIS_GH_APP_KEY_FILE` which is pointing to our pem file

Cheers!

xiao-pp:
@DanielWright I think you are right about #151 - I got confused too when I read the PR. However it does not stop one from setting it with Github App as both `ATLANTIS_GH_APP_ID` and `ATLANTIS_GH_APP_KEY_FILE` can be set via the variable `custom_environment_variables`? It's not ideal as it'd be nice if the module supports them as direct variables. But this PR is valid as it is indeed a bug there.

PR #171, opened 2020-12-11
feat: Allow define path_pattern, e.g. webhook /events bypass ALB auth…

jychen7:
…, but still have webhook secret validation

## Description
For webhook, not only restrict the IPs, but also restrict the path `/events` as well

## Motivation and Context
When `allow_unauthenticated_access=true`, the unauthenticated_access  IPs have access to `/events`, `/locks/:id` etc endpoints.
However, most use cases just require `/events` endpoint for unauthenticated_access (where it will validate webhook secret).

## Breaking Changes
nope, by default, it is same as original, because of dynamoc block

## How Has This Been Tested?
plan and applied


bryantbiggs:
hi @jychen7 - could you update your branch when you get a chance, thanks!

jychen7:
> could you update your branch when you get a chance, thanks!

@bryantbiggs sure, done in 05abd3c

README.md
```diff
@@ -144,7 +144,8 @@ If you are using one of the authentication methods above along with managed GitH
 
 ```hcl
 allow_unauthenticated_access = true
-allow_github_webhooks        = true
+allow_unauthenticated_access_path_pattern = "/events" # optional, by default is wildcard "*"
+allow_github_webhooks        = true # or manually set Github CIDRs in whitelist_unauthenticated_cidr_blocks
```
bryantbiggs, line 148: if this is set to true, it will add the GitHub CIDRS so this comment is not needed

main.tf
```diff
@@ -252,6 +252,12 @@ resource "aws_lb_listener_rule" "unauthenticated_access_for_cidr_blocks" {
     target_group_arn = module.alb.target_group_arns[0]
   }
 
+  condition {
```
bryantbiggs, line 255: while this may work (haven't fully tested the functionality) it does impose a change for all users. if we instead turned this into a dynamic block, we could instead have a no-op change for most users who don't opt into using an access path pattern

jychen7: right, `dynamic block` is better. will change it soon.

jychen7: fix in ea06460

jychen7:
@bryantbiggs would you mind review again? thanks

jychen7:
just update README from `master`

PR #172, opened 2020-12-18
fix: Update module dependencies for TF 0.14 support

emedvedev:
## Description

This PR updates module dependencies to provide support for Terraform 0.14.

## Motivation and Context

The versions of module dependencies that are used now set the `required_version` parameter as `< 0.14.0`, which means we cannot use the module with Terraform 0.14.

## Breaking Changes

Does not break compatibility: the new versions are all backwards-compatible.

## How Has This Been Tested?

Tested with `terraform validate` on Terraform 0.14.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -186,7 +186,7 @@ module "vpc" {
 ###################
 module "alb" {
   source  = "terraform-aws-modules/alb/aws"
-  version = "v5.7.0"
+  version = "v5.10.0"
 
   name     = var.name
   internal = var.internal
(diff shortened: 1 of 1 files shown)


bryantbiggs (approved):
@antonbabenko this is ready to merge - validated with complete example using `master` and then updated source to this change set. the only thing to note is that it looks like the security groups `Name` is wanting to update:

```hcl
 ~ tags = {
       "Environment" = "dev"
     ~ "Name"        = "atlantiscomplete-alb-https" -> "atlantiscomplete"
       "Owner"       = "user"
   }
```

antonbabenko:
Thanks @bryantbiggs for the review!

[v2.28.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.28.0) has been just released.

emedvedev:
Thank you!

PR #173, opened 2020-12-20
feat: allow adding more trusted principals to task role (users and ro…

kainlite:
## Description

This is adding another variable to be able to give some users or other roles the ability to assume the atlantis role, this can be refactored probably but I didn't want to add any breaking change for this simple requirement.

## Motivation and Context

While debugging it comes handy to be able to invoke terraform as atlantis does to speed things up, also this is particularly useful when importing resources.

## Breaking Changes

Optional variable, it should be a no op if empty.

## How Has This Been Tested?

Tested referencing my own branch.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -164,7 +164,7 @@ resource "aws_ssm_parameter" "atlantis_bitbucket_user_token" {
 ###################
 module "vpc" {
   source  = "terraform-aws-modules/vpc/aws"
-  version = "v2.47.0"
+  version = "v2.64.0"
 
   create_vpc = var.vpc_id == ""
 
@@ -186,7 +186,7 @@ module "vpc" {
 ###################
 module "alb" {
   source  = "terraform-aws-modules/alb/aws"
-  version = "v5.7.0"
+  version = "v5.10.0"
 
   name     = var.name
   internal = var.internal
@@ -264,7 +264,7 @@ resource "aws_lb_listener_rule" "unauthenticated_access_for_cidr_blocks" {
 ###################
 module "alb_https_sg" {
   source  = "terraform-aws-modules/security-group/aws//modules/https-443"
-  version = "v3.15.0"
+  version = "v3.17.0"
 
   name        = "${var.name}-alb-https"
   vpc_id      = local.vpc_id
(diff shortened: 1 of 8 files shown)


parmouraly:
Hey @kainlite 
I was just about to open an unrelated PR and came across this one here. 
Additional trusted principals are already supported
https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/master/variables.tf#L242-L246

kainlite:
> Hey @kainlite
> 
> Additional trusted principals are already supported from release [2.27.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/master/CHANGELOG.md#v2270---2020-12-10)
> 
> https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/master/variables.tf#L242-L246

Yes this is very similar, but the `trusted_principals` is only for AWS Services, my PR gives you the possibility to allow users or other roles to assume the atlantis role, this is particularly useful when importing, so your state get's the right permissions, owner, etc and you don't need to do any additional patching afterwards.

parmouraly:
> Yes this is very similar, but the `trusted_principals` is only for AWS Services, my PR gives you the possibility to allow users or other roles to assume the atlantis role, this is particularly useful when importing, so your state gets the right permissions, owner, etc and you don't need to do any additional patching afterwards.

Got you, makes sense 👍 
I don't perform imports via Atlantis but I guess if you are then I can see the value.

kainlite:
Updated from master

OneBadSanta:
I need this functionality as well. This has been sitting for a while. Is there any technical reason why this has not been merged?

main.tf
```diff
@@ -375,6 +375,11 @@ data "aws_iam_policy_document" "ecs_tasks" {
       type        = "Service"
       identifiers = compact(distinct(concat(["ecs-tasks.amazonaws.com"], var.trusted_principals)))
     }
+
+    principals {
+      type        = "AWS"
+      identifiers = compact(distinct(var.trusted_entities))
```
antonbabenko, line 381: This code won't work if `trusted_entities` is empty. Please rewrite it to use `dynamic`.

antonbabenko (changes requested):

main.tf
OneBadSanta: That is indeed a good reason :) Thanks for the quick response. I will rewrite it

OneBadSanta: @antonbabenko I wasn't sure how to contribute to this PR so i created another. Hope that's OK.

https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/193

antonbabenko:
Please use [v2.37.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.37.0) which has been just released (#193).

PR #189, opened 2021-03-24
fix: Allows users to set enable_dns_hostnames in VPC

zscholl:
## Description
This change adds a boolean variable called `enable_dns_hostnames` that gets passed into the module and used with the VPC module to set `enable_dns_hostnames` in the VPC. 

The default value for the variable is `false` which preserves the current default behavior of setting `enable_dns_hostnames` to `false` in the VPC.

## Motivation and Context
In order to resolve hostnames in a private Route53 hosted zone, the VPC must have `enable_dns_hostnames` set to `true`.

See AWS documentation here for [reference](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html#vpc-private-hosted-zones)
Screen Shot 2021-03-24 at 4 24 01 PM

We recently deployed Atlantis and are using it to make changes on infrastructure that has hostnames saved in a private Route53 hosted zone in AWS. In order for Atlantis to resolve these hostnames in Fargate we had to associate the private hosted zone with the VPC, and set `enable_dns_hostnames` to `true`. I am proposing this change so that I and others can modify this setting for this situation without manually editing the VPC or using a forked provider.

## Breaking Changes
N/A -- the default behavior will not be affected with this change.

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,167 +177,172 @@ allow_github_webhooks        = true
 
 | Name | Version |
 |------|---------|
-| terraform | >= 0.12.26 |
-| aws | >= 2.68 |
-| random | >= 2.0 |
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.12.26 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 2.68 |
+| <a name="requirement_random"></a> [random](#requirement\_random) | >= 2.0 |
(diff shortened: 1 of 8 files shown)


bryantbiggs:
as discussed in #202, users can elect to use their own VPC created outside of this module, but we will not continue adding in functionality to match that of the external VPC module

PR #192, opened 2021-04-04
feat: mdawn ecs policy gov compatibility

mdawn:
## Description

RIght now, the default partition is hard-coded to `aws`, this leverages the data source `aws_partition` https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/partition allowing this module to support GovCloud out of the box.

Thanks to @esacteksab for the original inspo to make it compatible.

## Motivation and Context
The module out of the box does not support GovCloud. This allows it to.

## Breaking Changes
None that I'm aware of.

## How Has This Been Tested?
Made this change on a local branch and ran it against our GovCloud instance. I see no reason why this would impact anything else.


antonbabenko:
Hi,

I believe you can just override `policies_arn` argument with the list of ARNs you have in GovCloud or any other part of AWS (eg, China).

```
policies_arn = ["arn:aws-us-gov:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"]
```

There is no need for this pull request.

PR #195, opened 2021-04-08
feat: Add enable deletion protection and drop invalid header fields

OneBadSanta:
## Description
This PR allows for configuring the enable_deletion_protection and drop_invalid_header_fields for the ALB. Its variables default to false, which is the default ALB module.

## Motivation and Context
I needed to configure these options for compliance reasons.

## Breaking Changes
No

## How Has This Been Tested?
Tested on own infrastructure


OneBadSanta:
@antonbabenko Are there any additional changes I need to make or does what is submitted look sufficient? Thx

examples/github-complete/variables.tf
```diff
@@ -14,6 +14,18 @@ variable "alb_ingress_cidr_blocks" {
   type        = list(string)
 }
 
+variable "alb_enable_deletion_protection" {
```
bryantbiggs, line 17: you can remove these two variables from the example variables

examples/github-complete/main.tf
```diff
@@ -81,6 +81,8 @@ module "atlantis" {
   alb_log_bucket_name             = module.atlantis_access_log_bucket.this_s3_bucket_id
   alb_log_location_prefix         = "atlantis-alb"
   alb_listener_ssl_policy_default = "ELBSecurityPolicy-TLS-1-2-2017-01"
+  alb_enable_deletion_protection  = var.alb_enable_deletion_protection
```
bryantbiggs, line 84: lets remove this line from the example

examples/github-complete/main.tf
```diff
@@ -81,6 +81,8 @@ module "atlantis" {
...
+  alb_drop_invalid_header_fields  = var.alb_drop_invalid_header_fields
```
bryantbiggs, line 85: lets set this line to `true`:
```suggestion
  alb_drop_invalid_header_fields  = true
```

examples/github-complete/variables.tf
```diff
@@ -14,6 +14,18 @@ variable "alb_ingress_cidr_blocks" {
...
+  description = "If true, deletion of the load balancer will be disabled via the AWS API. This will prevent Terraform from deleting the load balancer. Defaults to false."
+  type        = bool
+  default     = false
```
bryantbiggs, line 20: lets set these to `null` for defaults to rely on whatever is setup upstream or by the API
```suggestion
  default     = null
```

examples/github-complete/variables.tf
```diff
@@ -14,6 +14,18 @@ variable "alb_ingress_cidr_blocks" {
...
+}
+
+variable "alb_drop_invalid_header_fields" {
+  description = "Indicates whether invalid header fields are dropped in application load balancers. Defaults to false."
+  type        = bool
+  default     = false
```
bryantbiggs, line 26:
```suggestion
  default     = null
```

variables.tf
```diff
@@ -120,6 +120,18 @@ variable "alb_authenticate_cognito" {
   default     = {}
 }
 
+variable "alb_enable_deletion_protection" {
+  description = "If true, deletion of the load balancer will be disabled via the AWS API. This will prevent Terraform from deleting the load balancer. Defaults to false."
+  type        = bool
+  default     = false
```
bryantbiggs, line 126: lets set these to default to `null` so that they default to what is provided upstream by the provider or by the AWS API:
```suggestion
  default     = null
```

variables.tf
```diff
@@ -120,6 +120,18 @@ variable "alb_authenticate_cognito" {
...
+}
+
+variable "alb_drop_invalid_header_fields" {
+  description = "Indicates whether invalid header fields are dropped in application load balancers. Defaults to false."
+  type        = bool
+  default     = false
```
bryantbiggs, line 132:
```suggestion
  default     = null
```

OneBadSanta:
@bryantbiggs Changes sound good to me. Please see updates

bryantbiggs (approved):
@antonbabenko 👍🏼

OneBadSanta:
Thanks @bryantbiggs

variables.tf
```diff
@@ -120,6 +120,18 @@ variable "alb_authenticate_cognito" {
   default     = {}
 }
 
+variable "alb_enable_deletion_protection" {
+  description = "If true, deletion of the load balancer will be disabled via the AWS API. This will prevent Terraform from deleting the load balancer. Defaults to false."
+  type        = bool
+  default     = null
```
antonbabenko, line 126: These two variables should have default `false` since this is what the ALB module has as defaults - https://github.com/terraform-aws-modules/terraform-aws-alb/blob/master/variables.tf#L7-L17

Btw, description is correctly saying `false` :)

If we change this to `null` it will require an update for everyone using this module.

bryantbiggs: the ALB module makes assumptions by setting `false` as default, but the underlying provider/API ultimately defaults to `false` as well. I was trying to avoid making any assumptions with defaults if they are not required:
image
image

OneBadSanta: @antonbabenko  Personally, I can go either way. However, what @bryantbiggs suggested made sense to me in that setting the values to null is relying on AWS to provide these reasonable or possible preferable defaults. Setting them to default as false in this modules opinionates it further (which may or may not be preferred). 

I'm not entirely clear on how setting this to null will require everyone to update, but that does not sound like a good time.

Like I mentioned, I can go either way. I have only been using this module a couple weeks so I defer to the maintainers judgement. Please tell me what changes to make and I will make them. Thanks

antonbabenko: I misread all this. You are right, it should be `null`, my bad 👍  Merging right away!

antonbabenko:
Thank you guys for this PR!

[v2.38.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v2.38.0) has been just released.

PR #190, opened 2021-03-26
Adds support for aws_partition allowing for GovCloud support

esacteksab:
## Description
RIght now, the default partition is hard-coded to `aws`, this leverages the data source `aws_partition` https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/partition allowing this module to support GovCloud out of the box. 

## Motivation and Context
The module out of the box does not support GovCloud. This allows it to. 

## Breaking Changes
None that I'm aware of.

## How Has This Been Tested?
- [ ] I have tested and validated these changes using one or more of the provided `examples/*` projects

Made this change on a local branch and ran it against our GovCloud instance. I see no reason why this would impact anything else.

PR #202, opened 2021-04-25
refactor: remove VPC module from source

bryantbiggs:
## Description
- remove embedded VPC module from within the Atlantis module
- add data source for using correct partition based on where Atlantis is provisioned for the default ECS task execution policy ARN
- update sub-modules to use latest versions; GitHub version need to be updated for deprecating changes
- correct `github-complete` example project so it works correctly agin (trusted principals and trusted entities)

## Motivation and Context
- We already have a great VPC module and I don't see a reason to bake it into this module to further complicate this one. I think the better approach is to let users manage the VPC separately from this module to reduce the complexity managed here

- Closes #189 
- Closes #190 
- Closes #192
- Closes #201 

## Breaking Changes
- Yes; because the embedded VPC module is being removed and with the updates for the sub-module versions this should be considered a breaking change. Because the option was always present to use an existing VPC, I don't know if an upgrade guide is required - please let me know and I can create one (users who are creating a VPC from within this module will need to re-create no-matter what or step through several state move executions. Atlantis is stateless so its simpler to re-create IMO)

## How Has This Been Tested?
	- tested by updating/correcting the `github-complete` example

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: git://github.com/antonbabenko/pre-commit-terraform
-    rev: v1.48.0
+    rev: v1.50.0
     hooks:
       - id: terraform_fmt
       - id: terraform_validate
--- a/README.md
+++ b/README.md
@@ -192,15 +192,14 @@ allow_github_webhooks        = true
 
 | Name | Source | Version |
 |------|--------|---------|
-| <a name="module_acm"></a> [acm](#module\_acm) | terraform-aws-modules/acm/aws | v2.12.0 |
-| <a name="module_alb"></a> [alb](#module\_alb) | terraform-aws-modules/alb/aws | v5.10.0 |
-| <a name="module_alb_http_sg"></a> [alb\_http\_sg](#module\_alb\_http\_sg) | terraform-aws-modules/security-group/aws//modules/http-80 | v3.17.0 |
-| <a name="module_alb_https_sg"></a> [alb\_https\_sg](#module\_alb\_https\_sg) | terraform-aws-modules/security-group/aws//modules/https-443 | v3.17.0 |
-| <a name="module_atlantis_sg"></a> [atlantis\_sg](#module\_atlantis\_sg) | terraform-aws-modules/security-group/aws | v3.17.0 |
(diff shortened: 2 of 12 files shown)


antonbabenko:
What is the reason to remove VPC module from here?

I used this module several times as is in my projects and I liked the fact that I call just one module and it creates its own VPC used just by Atlantis.

If some users don't want to create a new VPC but prefer to manage it externally they can specify `vpc_id` and other required properties.

I really prefer to keep this module embedded as an option.

Lmk what you think.

bryantbiggs:
hey @antonbabenko, apologies I overlooked the comment. in my opinion I don't know that it makes a lot of sense to have this module package its own VPC, but our other modules do not. generally VPCs are quite sensitive and carefully thought out (I've personally deployed quite a few Atlantis instances using this module, but never once used the built-in VPC because thats usually crafted and controlled elsewhere). Its also another object that comes with a set of dependencies, users will want to add/modify attributes as the underlying module changes, etc. 

My goal was to:
- Package up the known breaking changes here in one step: remove VPC sub-module, bump to 0.13.x, bump underlying modules to latest which removes `this_*`, etc.
- Next step was to refactor the ALB portion: I don't know yet which makes more sense but either remove the ALB sub-module and provide just the necessary components to make it easy to attach to an ALB (again, relying on our ALB module which is provisioned outside of this module) - or - make it option to use the built-in ALB sub-module or provider your own ALB (more complex, complicated)

antonbabenko:
@bryantbiggs I am very busy with other things for the last few days. I will try to reply&review during the weekend. Sorry for that.

antonbabenko:
The reason why this module uses the VPC module and other our terraform-aws-modules don't do this is that this module can be used as a standalone component (aka, [infrastructure module](https://www.terraform-best-practices.com/key-concepts#infrastructure-module)). As soon as users want to customize any argument not supported by the module transparently they have an option to bring the VPC module (or some other module we use inside) next to this one and use it. Let's keep this behavior since it won't require a lot of work but rather useful to bring Atlantis up and running in some form.

I agree with all other changes you mention - upgrades of deps, 0.13+.

ALB... `aws_lb_listener_rule` is supported by the module, so it will be possible to use the ALB module for that. I don't think we need to extract ALB outside of the module.

nitrocode:
We could also say that any additional changes to the VPC, just disable the VPC management by this module, create the VPC outside of this module, and reference the outside VPC in this module. Best of both worlds.

PR #213, opened 2021-07-14
SRE 2508 update from upstream

raizyr:
- Add EFS volume mount
- Use volume name -data to match atlantis-data mount point
- Allow for authorization_config to be passed


raizyr:
This was supposed to be pushed to our fork, not upstream.  Closing it.

PR #217, opened 2021-08-18
allow mixing IPV4 and IPV6 CIDR blocks for the AWS ALB

velimir:
### Description

Allow mixing IPV4 and IPV6 CIDR blocks for the AWS ALB.

### Motivation and Context

Github started using IPv6 addresses.

If Atlantis is configured to allow GitHub IPs (via `alb_ingress_cidr_blocks` using "github" provider), terraform would fail with the following error:

``` 
Error: Error authorizing security group rule type ingress: InvalidParameterValue: CIDR block 2606:50c0::/32 is malformed
```

### Breaking Changes

No breaking changes.

### How Has This Been Tested?

Tested manually using mixed range of IPv6 and IPv4 addresses.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -298,7 +298,8 @@ module "alb_http_sg" {
   vpc_id      = local.vpc_id
   description = "Security group with HTTP ports open for specific IPv4 and IPv6CIDR block (or everybody), egress ports are all world open"
 
-  ingress_cidr_blocks = sort(compact(concat(var.allow_github_webhooks ? var.github_webhooks_cidr_blocks : [], var.alb_ingress_cidr_blocks)))
+  ingress_cidr_blocks = local.alb_ingress_ipv4_cidr_blocks
+  ingress_ipv6_cidr_blocks = local.alb_ingress_ipv6_cidr_blocks
 
   tags = merge(local.tags, var.alb_http_security_group_tags)
 }
--- a/main.tf
+++ b/main.tf
@@ -96,6 +96,13 @@ locals {
     },
   ] : []
 
+  alb_ingress_cidr_blocks = sort(compact(concat(var.allow_github_webhooks ?   var.github_webhooks_cidr_blocks : [], var.alb_ingress_cidr_blocks)))
+  alb_ingress_ipv6_cidr_blocks =  [
+    for cidr in local.alb_ingress_cidr_blocks : cidr
+    if length(split(":", cidr)) > 1
+  ]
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -288,9 +296,10 @@ module "alb_http_sg" {
 
   name        = "${var.name}-alb-http"
   vpc_id      = local.vpc_id
-  description = "Security group with HTTP ports open for specific IPv4 CIDR block (or everybody), egress ports are all world open"
+  description = "Security group with HTTP ports open for specific IPv4 and IPv6CIDR block (or everybody), egress ports are all world open"
 
-  ingress_cidr_blocks = sort(compact(concat(var.allow_github_webhooks ? var.github_webhooks_cidr_blocks : [], var.alb_ingress_cidr_blocks)))
+  ingress_cidr_blocks = local.alb_ingress_ipv4_cidr_blocks
+  ingress_ipv6_cidr_blocks = local.alb_ingress_ipv6_cidr_blocks
```
antonbabenko, line 302: Please introduce another variable like `var.alb_ingress_ipv6_cidr_blocks` and use it instead of this unnecessary magic :)

velimir: sure, will do 👍

velimir: @antonbabenko so... the thing is the magic in the locals takes into account blocks from 2 lists `var.allow_github_webhooks` and `var.alb_ingress_cidr_blocks`.

Do you think there should be ipv6 list of the GitHub web hooks blocks accordingly?

antonbabenko: `var.github_webhooks_cidr_blocks` (right?) contains IPv4 addresses so no need to do anything with it.

velimir: yes, sorry the `var.github_webhooks_cidr_blocks`.

At the moment it can only accept IPv4 (since it's passed to the `ingress_cidr_blocks` of the security group rule.

Since now GitHub started to use IPv6 for web hooks (`2a0a:a440::/29` and `2606:50c0::/32`), how do we pass these addresses to the security group rule as well? Those can't go in the same `var.github_webhooks_cidr_blocks` variable as that one is limited to IPv4 only.

antonbabenko: Then let's introduce another variable for GitHub IPs like `var.github_webhooks_ipv6_cidr_blocks` and have them separated.

antonbabenko:
This issue has been resolved in version 3.15.0 :tada:

PR #219, opened 2021-08-31
feat: update default image to ghcr.io one

chenrui333:
## Description
per [Atlantis 0.17.1 release](https://github.com/runatlantis/atlantis/releases/tag/v0.17.1), the default image registry is ghcr.io. Thus updating the code to reflect that.

### DIFF
--- a/examples/github-complete/main.tf
+++ b/examples/github-complete/main.tf
@@ -52,7 +52,7 @@ module "atlantis" {
   docker_labels = {
     "org.opencontainers.image.title"       = "Atlantis"
     "org.opencontainers.image.description" = "A self-hosted golang application that listens for Terraform pull request events via webhooks."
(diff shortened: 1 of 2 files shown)


antonbabenko:
Thanks @chenrui333 !

[v3.1.0](https://github.com/terraform-aws-modules/terraform-aws-atlantis/releases/v3.1.0) has been just released.

chenrui333:
Thanks @antonbabenko!

bryantbiggs:
FYI - I don't think this change does what you think it does. if we wanted to change the image repo, we should have changed the image version to be `dev` per the directions of upstream Atlantis repo https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/179b846000b64d8e7e42890c5672455b4db9bf3b/variables.tf#L470

bryantbiggs:
also, they are still publishing to Dockerhub as well so the "old" route still works

PR #233, opened 2021-11-03
feat: Allow for an existing ECS cluster to be used

henworth:
## Description
Provide the ability to use an existing ECS cluster instead of making one.

## Motivation and Context
Allow for the use of an existing cluster if there is a desire or requirement to do so.

## Breaking Changes
None.

## How Has This Been Tested?

Ran this in my own environment to build my Atlantis instance using my existing cluster. This also passes all of the pre-commit tests.


main.tf
```diff
@@ -379,6 +383,7 @@ resource "aws_route53_record" "atlantis" {
 # ECS
 ################################################################################
 module "ecs" {
+  count   = var.ecs_cluster_id == "" ? 1 : 0
```
bryantbiggs, line 386: this module has a conditional that should be used instead of count - [`create_ecs`](https://github.com/terraform-aws-modules/terraform-aws-ecs#conditional-creation)

henworth: Nice! Did not know this.

outputs.tf
```diff
@@ -53,12 +53,12 @@ output "ecs_security_group" {
...
 output "ecs_cluster_id" {
   description = "ECS cluster id"
-  value       = module.ecs.ecs_cluster_id
+  value       = local.ecs_cluster_id
 }
 
 output "ecs_cluster_arn" {
   description = "ECS cluster ARN"
-  value       = module.ecs.ecs_cluster_arn
+  value       = local.ecs_cluster_arn
```
bryantbiggs, line 61: according to the docs, both ID and ARN are the same (ARN). @antonbabenko what if we just set these both to `value = local.ecs_cluster_id`? or should we leave this as `module.ecs.cluster_arn` knowing it will be empty when you don't create a cluster with this module?

either way, we should drop the `ecs_cluster_arn` variable, its not providing value here since we are asking for the ID. lets remove that variable and the associated local variable, thanks!

henworth: Happy to remove it, I was just maintaining existing compatibility.

antonbabenko: +1 @bryantbiggs as long as it works for both cases. `local.ecs_cluster_id` sounds valid to me.

outputs.tf
```diff
@@ -53,12 +53,7 @@ output "ecs_security_group" {
 
 output "ecs_cluster_id" {
   description = "ECS cluster id"
-  value       = module.ecs.ecs_cluster_id
-}
-
-output "ecs_cluster_arn" {
```
bryantbiggs, line 59: lets leave this in for now so its not technically a "breaking change" - it will just be empty when the cluster is not created by the module. we can update the description to reflect with "ID of ECS cluster created"

bryantbiggs:
everything checks out for this PR - if we can just address the one detail listed above we should be good to go - thanks @henworth  and apologies for the delay

henworth:
@bryantbiggs no problem, I've reverted that output removal

bryantbiggs:
@henworth looks like you need to run `pre-commit run -a` one last time to fix the docs - otherwise should be good to go @antonbabenko

henworth:
oops, fixed @bryantbiggs

bryantbiggs (approved):
thanks for the PR @henworth - good to go @antonbabenko 👍🏽

antonbabenko:
This PR is included in version 3.12.0 :tada:

PR #234, opened 2021-11-04
docs: Added google oidc example to readme

devdavidkarlsson:
## Description

We are using this with google oidc

## Motivation and Context
Having google OIDC as an example makes sense for us: Especially since some may forget to configure their google consent screen to internal, that note could be worth mentioning.

## Breaking Changes
Only readme example added

## How Has This Been Tested?
We run this example our selves in our envs.


antonbabenko:
Thank you very much for this addition!

PR #235, opened 2021-11-10
feat: bring your own ECS + LB

endrec:
## Description
A set of opinionated changes to allow one to use their existing ECS and LB.

## Motivation and Context
We wanted to deploy multiple atlantis servers (one per environment), but use the same ECS cluster and load balancer.
This change depends on https://github.com/terraform-aws-modules/terraform-aws-alb/pull/227.

## Breaking Changes
Removed input_allow_repo_config variable, as the setting is deprecated in atlantis.

## How Has This Been Tested?
We run these changes ourselves.


main.tf
```diff
@@ -192,13 +196,26 @@ module "vpc" {
...
+data "aws_lb_listener" "https_listener" {
+  load_balancer_arn = local.alb_arn
+  port              = 443
+}
+
 module "alb" {
-  source  = "terraform-aws-modules/alb/aws"
-  version = "v6.5.0"
+  source = "git@github.com:endrec/terraform-aws-alb.git?ref=byo"
+  #  source  = "terraform-aws-modules/alb/aws"
```
bryantbiggs, line 211: we should leave this as the ALB module managed under this organization

bryantbiggs (changes requested):
we will need to stick with the ALB module provided under the `terraform-aws-modules` organization

main.tf
endrec: Sure we should, but I don't know the version number till terraform-aws-modules/terraform-aws-alb#227 is released.

endrec:
This PR depends on a change which won't get merge: 
https://github.com/terraform-aws-modules/terraform-aws-alb/pull/227#issuecomment-1009891950

PR #244, opened 2022-01-17
feat: Add base URL for GitHub Enterprise

DoomHammer:
## Description

Add base URL for use with GitHub Enterprise.

## Motivation and Context

Currently, the only way to use the module with GItHub Enterprise is via injecting the provider configuration. However, in recent versions of Terraform, `terraform validate` will treat it as an error. Adding a variable solves this issue.

## Breaking Changes

No known.

## How Has This Been Tested?

### DIFF
--- a/modules/github-repository-webhook/variables.tf
+++ b/modules/github-repository-webhook/variables.tf
@@ -7,7 +7,7 @@ variable "create_github_repository_webhook" {
 variable "github_base_url" {
   description = "Github base URL to use when creating webhook (when using GitHub Enterprise)"
   type        = string
-  default     = ""
+  default     = null
 }
 
 variable "github_token" {
--- a/modules/github-repository-webhook/variables.tf
+++ b/modules/github-repository-webhook/variables.tf
@@ -4,6 +4,12 @@ variable "create_github_repository_webhook" {
   default     = true
 }
 
+variable "github_base_url" {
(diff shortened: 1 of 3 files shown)


modules/github-repository-webhook/variables.tf
```diff
@@ -4,6 +4,12 @@ variable "create_github_repository_webhook" {
   default     = true
 }
 
+variable "github_base_url" {
+  description = "Github base URL to use when creating webhook (when using GitHub Enterprise)"
+  type        = string
+  default     = ""
```
antonbabenko, line 10:
```suggestion
  default     = null
```

Accoding to [the docs](https://registry.terraform.io/providers/integrations/github/latest/docs#base_url) for `base_url` it is optional, so it should be `null`.

antonbabenko (commented):
LGTM, but just a minor comment.

modules/github-repository-webhook/variables.tf
DoomHammer: Ah, sorry about that and thanks for suggestion!

antonbabenko: Also, please fix the docs (to make CI green).

antonbabenko:
This PR is included in version 3.7.0 :tada:

PR #247, opened 2022-01-30
Persist Atlantis state using EFS

### LINKED ISSUE #206: Leverage EFS to persist atlantis locks between deployments?

Every time we make a change that requires redeploying our atlantis container, we lose all of the locks on PRs that atlantis currently has. These locks are persisted to the file system which is not maintained between containers.

## Describe the solution you'd like.
I'm wondering if the module could configure an EFS mount where the file system locks can be maintained, so that they persist between deployments.

## Describe alternatives you've considered.
Currently we have to capture the current list of locks, deploy atlantis, and then go to each of the PRs to reacquire the locks.

MarkIannucci:
## Description
Use EFS to store Atlantis state (we are most concerned about locks, but this also reduces repo checkouts in some cases).

## Motivation and Context
Fixes #206

## Breaking Changes
Reduces the flexibility we offer by requiring the user variable to be in the `uid:gid` format.  This is because I wanted to use the user variable for the `posix_user` defined in the `aws_efs_access_point` and I wasn't aware of a method to decompose the more flexible entries we allowed previously into their uid and gid parts.  If someone is aware of a more elegant way to do that, I look forward to learning from your contribution!  In the meantime, I added a validation to the variable to surface the breaking change at plantime rather than after the deployment.

I also change the default maximum deployment percent and minimum healthy deployment percent to 100 and 0 respectively.  I made this change because I found that Atlantis gets angry when there are two instances sharing the same BoltDB database.  This means that we have brief interruptions when we make changes that require a new ECS deployment in AWS, but I think it is better to have the interruptions than lost locks.

## How Has This Been Tested?
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects --- I tested this with the github example.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -191,6 +191,7 @@ module "vpc" {
 
   enable_nat_gateway = true
   single_nat_gateway = true
+  enable_dns_hostnames = true
 
   tags = local.tags
 }
@@ -351,6 +352,27 @@ module "atlantis_sg" {
   tags = merge(local.tags, var.atlantis_security_group_tags)
 }
 
+module "efs_sg" {
+  source = "terraform-aws-modules/security-group/aws"
+  version = "v4.8.0"
+
+  name = "${var.name}-efs"
+  vpc_id = local.vpc_id
+  description = "Security group allowing access to the EFS storage"
+
+  ingress_with_source_security_group_id = [
+    {
+      from_port = 2049
+      to_port = 2049
+      protocol = "tcp"
+      description = "nfs for EFS"
+      source_security_group_id = module.atlantis_sg.security_group_id
+    }
+  ]
+
+  tags = local.tags
+}
+
 ################################################################################
 # ACM (SSL certificate)
 ################################################################################
@@ -381,6 +403,30 @@ resource "aws_route53_record" "atlantis" {
     name                   = module.alb.lb_dns_name
     zone_id                = module.alb.lb_zone_id
     evaluate_target_health = true
+  }
+}
+
+################################################################################
+# EFS
+################################################################################
+
+resource "aws_efs_file_system" "efs" {
+  creation_token = var.name
+}
+
+resource "aws_efs_mount_target" "efs-mt" {
+  # didn't use foreach because terraform doesn't know how many subnets will exist on initial create
+  count = length(local.private_subnet_ids)
+  file_system_id = aws_efs_file_system.efs.id
+  subnet_id = tolist(local.private_subnet_ids)[count.index]
+  security_groups = [module.efs_sg.security_group_id, module.atlantis_sg.security_group_id]
+}
+
+resource "aws_efs_access_point" "efs-ap" {
+  file_system_id = aws_efs_file_system.efs.id
+  posix_user {
+    gid = 0
+    uid = 0
   }
 }
 
@@ -624,6 +670,19 @@ resource "aws_ecs_task_definition" "atlantis" {
       size_in_gib = var.ephemeral_storage_size
     }
   }
+
+  volume {
+     name = "efs-storage"
+     efs_volume_configuration {
+      file_system_id = aws_efs_file_system.efs.id
+      transit_encryption      = "ENABLED"
+      transit_encryption_port = 2999
+      authorization_config {
+        access_point_id = aws_efs_access_point.efs-ap.id
+        iam             = "ENABLED"
+      }
+     }
+   }
 
   tags = local.tags
 }
--- a/main.tf
+++ b/main.tf
@@ -110,6 +110,17 @@ locals {
     sort(compact(concat(var.allow_github_webhooks ? var.github_webhooks_cidr_blocks : [], var.whitelist_unauthenticated_cidr_blocks))),
     5
   )
+
+  # break up user to uid and gid -- set both to 0 if null
+  uid = var.user == null ? 0 : split(":", var.user)[0]
+  gid = var.user == null ? 0 : split(":", var.user)[1]
+
+  # default mount points for efs if ephemeral storage is not enabled and mount points aren't specified
+  mount_points = var.enable_ephemeral_storage ? var.mount_points : length(var.mount_points) > 0 ? var.mount_points : [{
+      containerPath = "/home/atlantis"
+      sourceVolume = "efs-storage"
+      readOnly = "false"
+    }]
 }
 
 data "aws_partition" "current" {}
@@ -353,6 +364,7 @@ module "atlantis_sg" {
 }
 
 module "efs_sg" {
+  count = var.enable_ephemeral_storage ? 0 : 1
   source = "terraform-aws-modules/security-group/aws"
   version = "v4.8.0"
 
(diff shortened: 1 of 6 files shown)


main.tf
```diff
@@ -110,6 +110,17 @@ locals {
     sort(compact(concat(var.allow_github_webhooks ? var.github_webhooks_cidr_blocks : [], var.whitelist_unauthenticated_cidr_blocks))),
     5
   )
+
+  # break up user to uid and gid -- set both to 0 if null
+  uid = var.user == null ? 0 : split(":", var.user)[0]
+  gid = var.user == null ? 0 : split(":", var.user)[1]
+
+  # default mount points for efs if ephemeral storage is not enabled and mount points aren't specified
+  mount_points = var.enable_ephemeral_storage ? var.mount_points : length(var.mount_points) > 0 ? var.mount_points : [{
```
antonbabenko, line 119:
```suggestion
  mount_points = var.enable_ephemeral_storage || length(var.mount_points) > 0 ? var.mount_points : [{
```

main.tf
```diff
@@ -191,6 +202,7 @@ module "vpc" {
 
   enable_nat_gateway = true
   single_nat_gateway = true
+  enable_dns_hostnames = true
```
antonbabenko, line 205: I think this should be conditional, like this:

```suggestion
  enable_dns_hostnames = var.enable_ephemeral_storage
```

main.tf
```diff
@@ -355,6 +367,28 @@ module "atlantis_sg" {
   tags = merge(local.tags, var.atlantis_security_group_tags)
 }
 
+module "efs_sg" {
+  count = var.enable_ephemeral_storage ? 0 : 1
```
antonbabenko, line 371: Move line with `count` after `version`, and separate them with a newline.

main.tf
```diff
@@ -355,6 +367,28 @@ module "atlantis_sg" {
...
+  source = "terraform-aws-modules/security-group/aws"
+  version = "v4.8.0"
+
+  name = "${var.name}-efs"
+  vpc_id = local.vpc_id
+  description = "Security group allowing access to the EFS storage"
+
+  ingress_with_source_security_group_id = [
```
antonbabenko, line 379: There is a separate submodule for `nfs/efs` - https://github.com/terraform-aws-modules/terraform-aws-security-group/tree/master/modules/nfs ([usage example for submodules](https://github.com/terraform-aws-modules/terraform-aws-security-group/blob/master/examples/http/main.tf)).

It will make your code a few lines shorter.

main.tf
```diff
@@ -388,6 +422,32 @@ resource "aws_route53_record" "atlantis" {
   }
 }
 
+################################################################################
+# EFS
+################################################################################
+
+resource "aws_efs_file_system" "efs" {
+  count = var.enable_ephemeral_storage ? 0 : 1
```
antonbabenko, line 430: Add a new-line separator after `count` and `for_each` in all resources and blocks.

main.tf
```diff
@@ -388,6 +422,32 @@ resource "aws_route53_record" "atlantis" {
...
+  creation_token = var.name
+}
+
+resource "aws_efs_mount_target" "efs-mt" {
```
antonbabenko, line 434:
```suggestion
resource "aws_efs_mount_target" "this" {
```

main.tf
```diff
@@ -388,6 +422,32 @@ resource "aws_route53_record" "atlantis" {
...
+  # didn't use foreach because terraform doesn't know how many subnets will exist on initial create
+  count = var.enable_ephemeral_storage ? 0 : length(local.private_subnet_ids)
+  file_system_id = aws_efs_file_system.efs[0].id
+  subnet_id = tolist(local.private_subnet_ids)[count.index]
+  security_groups = [module.efs_sg[0].security_group_id, module.atlantis_sg.security_group_id]
+}
+
+resource "aws_efs_access_point" "efs-ap" {
```
antonbabenko, line 442:
```suggestion
resource "aws_efs_access_point" "this" {
```

main.tf
```diff
@@ -388,6 +422,32 @@ resource "aws_route53_record" "atlantis" {
   }
 }
 
+################################################################################
+# EFS
+################################################################################
+
+resource "aws_efs_file_system" "efs" {
```
antonbabenko, line 429:
```suggestion
resource "aws_efs_file_system" "this" {
```

main.tf
```diff
@@ -388,6 +422,32 @@ resource "aws_route53_record" "atlantis" {
...
+  count = var.enable_ephemeral_storage ? 0 : 1
+  creation_token = var.name
+}
+
+resource "aws_efs_mount_target" "efs-mt" {
+  # didn't use foreach because terraform doesn't know how many subnets will exist on initial create
```
antonbabenko, line 435: I think `count` will have the same limitation here. `for_each` should be preferred more when order is not important. Could you give it a try and update?

antonbabenko (changes requested):
Looks pretty good. Few comments mostly about code style to match the rest of the module(s) we have.

main.tf
MarkIannucci: `enable_ephemeral_storage` actually defaults to false, so I inversed the bool and converted this to a conditional as you suggested in [b6d233e](https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/247/commits/b6d233e66dffd92b13d3111ba66ba874b764a37e)

main.tf
MarkIannucci: Thank you for encouraging me to revisit that resource.  I'm not sure why I got stuck initially, however I was able to find a way to get it done with `for_each` :)

MarkIannucci:
@antonbabenko , Thanks for the thorough review.  We differed in one place on line `205` -- I left the comment unresolved to highlight it.  This is ready for a re-review and potential completion.

variables.tf
```diff
@@ -332,13 +332,13 @@ variable "ecs_service_platform_version" {
 variable "ecs_service_deployment_maximum_percent" {
   description = "The upper limit (as a percentage of the service's desiredCount) of the number of running tasks that can be running in a service during a deployment"
   type        = number
-  default     = 200
+  default     = 100
```
antonbabenko, line 335: Why do you change these defaults for ECS service? If this isn't critical for this PR which is about EFS, please revert.

antonbabenko (changes requested):
Looks pretty good. Just a comment about `default` values. Also, please make CI green by running `pre-commit run -a` and committing the changes.

variables.tf
MarkIannucci: I reduced these defaults because I found that Atlantis would throw errors complaining about the BoltDB being in use when ECS would spin up the second container during ECS deployments.  The errors occurred in latest version of the container, so by reducing these defaults, ECS stops the existing container before starting the next one which avoids BoltDB being in use by two containers simultaneously.

It does come at a cost of a brief outage whenever an ECS deployment occurs.  In my testing that was always less than 5 minutes.  

I suppose we could add a local to increase these values if someone is using ephemeral storage.  Personally though, I'd rather have the outage than the lost locks in #206 which is what this PR fixes.

antonbabenko: Thanks for the explanation.

antonbabenko (approved):

main.tf
```diff
@@ -388,6 +418,35 @@ resource "aws_route53_record" "atlantis" {
...
+  count = var.enable_ephemeral_storage ? 0 : 1
+
+  creation_token = var.name
+}
+
+resource "aws_efs_mount_target" "this" {
+  # we coalescelist in order to specify the resource keys when we create the subnets using the VPC or they're specified for us.  This works around the for_each value depends on attributes which can't be determined until apply error
+  for_each = zipmap(coalescelist(var.private_subnets, var.private_subnet_ids), local.private_subnet_ids)
+
+  file_system_id  = aws_efs_file_system.this[0].id
```
borissavelev, line 435: @MarkIannucci 
first of all, thank you! 🙏 

Unfortunately this is not working if I want to have `enable_ephemeral_storage=true`

```
╷
│ Error: Invalid index
│ 
│   on .terraform/modules/atlantis/main.tf line 438, in resource "aws_efs_mount_target" "this":
│  438:   file_system_id  = aws_efs_file_system.this[0].id
│     ├────────────────
│     │ aws_efs_file_system.this is empty tuple
│ 
│ The given key does not identify an element in this collection value.
╵
```

there is no condition for `var.enable_ephemeral_storage`

main.tf
```diff
@@ -355,6 +367,24 @@ module "atlantis_sg" {
...
+module "efs_sg" {
+  source  = "terraform-aws-modules/security-group/aws//modules/nfs"
+  version = "v4.8.0"
+  count   = var.enable_ephemeral_storage ? 0 : 1
+
+  name        = "${var.name}-efs"
+  vpc_id      = local.vpc_id
+  description = "Security group allowing access to the EFS storage"
+
+  ingress_cidr_blocks = [var.cidr]
```
borissavelev, line 379: `var.cidr` can be empty when is reusing existing VPC

main.tf
borissavelev: no problem! thank you)

main.tf
borissavelev: probably `ingress_cidr_blocks` is unnecessary here because later we have `ingress_with_source_security_group_id`

PR #251, opened 2022-02-02
feat: Add var.task_execution_session_duration for task execution role

itmustbejj:
## Description
Add `var.task_execution_session_duration` for task execution role.

image
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role

## Motivation and Context
The default max_session_duration for iam roles of 1 hour is insufficient for long-running terraform applies. This leads to runs breaking and failing to persist their state in s3, and often requires painful manual cleanup.

## Breaking Changes
None

## How Has This Been Tested?

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -688,3 +688,9 @@ variable "ephemeral_storage_size" {
     error_message = "The minimum supported value is 21 GiB and the maximum supported value is 200 GiB."
   }
 }
+
+variable "max_session_duration" {
+  description = "Maximum session duration (in seconds) for ecs task execution role. Default is 3600."
+  type        = number
+  default     = null
+}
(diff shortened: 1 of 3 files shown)


variables.tf
```diff
@@ -672,3 +672,14 @@ variable "ephemeral_storage_size" {
     error_message = "The minimum supported value is 21 GiB and the maximum supported value is 200 GiB."
   }
 }
+
+variable "task_execution_session_duration" {
```
bryantbiggs, line 676: lets do the following please:

```hcl
variable "max_session_duration" {
  description = "Maximum session duration (in seconds) that you want to set for the specified role. Default is 3600"
  type        = number
  default     = null
}
```

Note - you'll need to rename the variable above in `main.tf`

bryantbiggs (changes requested):

variables.tf
itmustbejj: @bryantbiggs thanks for the prompt feedback. That should be all set.

bryantbiggs:
sorry to hear your plans take that long - might want to think about splitting up that statefile 😅

itmustbejj:
We use terragrunt so the state files are already split up pretty judiciously. The specific use case here is around long-running sync jobs. I have ideas for the next refactor but you know how that goes :sweat_smile:

variables.tf
antonbabenko: Also, please fix the docs by running `pre-commit run -a` and committing the changes.

itmustbejj:
@antonbabenko @bryantbiggs apologies for the delay. I rebased my PR and updated the README via `pre-commit run -a`.

PR #255, opened 2022-02-14
Use data source for VPC CIDR block

### LINKED ISSUE #252: Error with EFS security group if using existing VPC
## Description

If using an existing VPC rather than let this module create it, it now generates this error:
```
│ Error: "" is not a valid CIDR block: invalid CIDR address: 
│ 
│   with module.atlantis.module.efs_sg[0].module.sg.aws_security_group_rule.ingress_rules[0],
│   on .terraform/modules/atlantis.efs_sg/main.tf line 70, in resource "aws_security_group_rule" "ingress_rules":
│   70:   cidr_blocks      = var.ingress_cidr_blocks
│ 
╵
```
This is due to this line:
https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/2df6bd6d5350d38e42ad9c63103c642f60293a43/main.tf#L382

`var.cidr` doesn't need to be set if you have the VPC already created and are passing it in with `var.vpc_id`.

## Versions

 - Terraform: 1.1.3

 - Provider(s): aws 3.74.2

 - Module: 3.12.0

### Code Snippet to Reproduce
Use this module with the `vpc_id` parameter to use an existing VPC.

## Expected behavior
It shouldn't be erroring, possibly should be using a datasource to get the CIDR block associated with the VPC referenced by `var.vpc_id`.

## Actual behavior
See above.

bodgit:
## Description
This PR adds a VPC data source to retrieve the CIDR block of the VPC which is used in computing the rules for the EFS security group.

## Motivation and Context
Fixes #252 

## How Has This Been Tested?
Using the latest released module version results in an error at plan stage when using an existing VPC. With this PR, the plan completes and shows the correct VPC CIDR in the security group rules.


bryantbiggs (commented):
shouldn't we just remove `ingress_cidr_blocks` all together? at minimum, only Atlantis needs access via NFS to EFS; this seems like additional access which can be better handled by accepting a list of additional security group IDs here https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/2df6bd6d5350d38e42ad9c63103c642f60293a43/main.tf#L440

bodgit:
I can make that change, I was just trying to do the minimum to get the code to compile which was fixing what was there. Let me adjust the PR.

bodgit:
I needed to add `auto_ingress_rules = []` so it doesn't try and create rules using a nonexistent CIDR block.

aairey (approved):

bodgit:
Not stale

bodgit:
@bryantbiggs @antonbabenko Any chance this can be merged?

main.tf
```diff
@@ -379,7 +379,8 @@ module "efs_sg" {
   vpc_id      = local.vpc_id
   description = "Security group allowing access to the EFS storage"
 
-  ingress_cidr_blocks = [var.cidr]
+  auto_ingress_rules = []
```
antonbabenko, line 382: Setting it like this disables the main purpose of this security group submodule which creates a predefined rule (for `nfs` in this case):

https://github.com/terraform-aws-modules/terraform-aws-security-group/blob/master/modules/nfs/auto_values.tf#L6-L10

antonbabenko (commented):
I believe this PR is not needed because the root problem was fixed in https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/261#issuecomment-1086032476

Please upgrade your code to use version 3.13.1 of this module.

PR #256, opened 2022-02-15
feat: Add IPv6 support (ALB and Route53 AAAA record)

bodgit:
## Description
This PR enables IPv6 support for the VPC, ALB and adds a AAAA Route53 record.

## Motivation and Context
I need to run Atlantis with IPv6 connectivity. I manage the VPC outside of this module so that already has IPv6 enabled so primarily I just needed to change the ALB to "dualstack" and add the AAAA record. I've added additional variables so if the VPC is managed by this module the same configuration can be achieved.

## Breaking Changes
None. The new `enable_ipv6` variable defaults to false so existing IPv4-only behaviour is preserved.

## How Has This Been Tested?
This will perhaps need a new example?


bodgit:
> I think we can add the AAAA record, but if users want to use IPv6 then I think they should use an externally created VPC module otherwise we're slowly going to pull in that entire module here

That's what I'm doing anyway. I was managing the AAAA record separately too, the only necessary feature really is to be able to set the ALB `ip_address_type` attribute to "dualstack".

Should I just reduce the PR to exposing an `alb_ip_address_type` variable and then make the AAAA record conditional on that being set to "dualstack"?

main.tf
```diff
@@ -234,6 +242,8 @@ module "alb" {
     prefix  = var.alb_log_location_prefix
   }
 
+  ip_address_type = var.enable_ipv6 ? "dualstack" : "ipv4"
```
bryantbiggs, line 245: ya lets just make this
```suggestion
  ip_address_type = var.alb_ip_address_type
```

And the variable default will be `"ipv4"` like the ALB module does

bryantbiggs:
> > I think we can add the AAAA record, but if users want to use IPv6 then I think they should use an externally created VPC module otherwise we're slowly going to pull in that entire module here
> 
> That's what I'm doing anyway. I was managing the AAAA record separately too, the only necessary feature really is to be able to set the ALB `ip_address_type` attribute to "dualstack".
> 
> Should I just reduce the PR to exposing an `alb_ip_address_type` variable and then make the AAAA record conditional on that being set to "dualstack"?

ya see the two comments above, and then we can just drop the VPC module changes from here

bryantbiggs (approved):
looks good to me - what do you think @antonbabenko

bodgit:
> Also, please update the docs (run `pre-commit run -a` and commit changes).

How do I run that?

variables.tf
```diff
@@ -688,3 +688,15 @@ variable "ephemeral_storage_size" {
     error_message = "The minimum supported value is 21 GiB and the maximum supported value is 200 GiB."
   }
 }
+
+variable "alb_ip_address_type" {
+  description = "The type of IP addresses used by the subnets for your load balancer"
```
antonbabenko, line 693: Add available options (`ipv4` or `dualstack` - for IPv6) in the description.

antonbabenko (changes requested):
Just a minor comment before merging.

variables.tf
bodgit: Done

bodgit:
@bryantbiggs @antonbabenko can this be reopened and merged please? I'm not aware of any outstanding issues with it

(I really hate stale bots)

antonbabenko:
@bodgit

antonbabenko:
This PR is included in version 3.14.0 :tada:

PR #257, opened 2022-02-15
fix: Encrypt EFS + add tags to EFS

borissavelev:
## Description

For sake of security EFS resource should be encrypted.
Add default tags (including `Name`) to EFS resource.


aldwyn (changes requested):
This should also optionally include `kms_key_id` for custom KMS keys.

ax-nmac:
+1 for this.  This is a non-starter for our org (and probably most orgs) without encryption, and since EFS is default now we are stuck using an older version of the module.

PR #258, opened 2022-02-15
fix: Secrets in task definition should be as ARN 

borissavelev:
## Description
ECS `secrets` `valueFrom` should be as ARN

## Motivation and Context
[ContainerDefinition spec](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#ContainerDefinition-secrets)

current implementation puts into `valueFrom` string with SSM parameter path and ECS task cannot start

## Breaking Changes
No

## How Has This Been Tested?
It was tested with 

```
module "test" {
  source = "git::https://github.com/borissavelev/terraform-aws-atlantis.git?ref=SSM_fixes"

  atlantis_github_user       = var.github_user
  atlantis_github_user_token = var.github_token
}
```

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -19,7 +19,7 @@ locals {
 
   # token
   secret_name_key        = local.has_secrets ? var.atlantis_gitlab_user_token != "" ? "ATLANTIS_GITLAB_TOKEN" : var.atlantis_github_user_token != "" ? "ATLANTIS_GH_TOKEN" : "ATLANTIS_BITBUCKET_TOKEN" : ""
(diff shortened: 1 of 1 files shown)


antonbabenko:
According to notes in the AWS docs mentioned in the description:

> If the Systems Manager Parameter Store parameter exists in the same Region as the task you are launching then you can use either the full ARN or name of the secret. If the parameter exists in a different Region then the full ARN must be specified.

I guess, we currently cover the use-case when the region is the same, so it works.

This PR does something wrong (e.g. drops gitlab).

borissavelev:
yes, you are right! thanks for looking into it

PR #259, opened 2022-02-16
fix: Use random_password resource instead of random_id

### LINKED ISSUE #158: Use random_password to generate webhook token
This module uses `random_id` to generate the webhook token. `random_id` is not considered a secret resource.

Background: https://github.com/hashicorp/terraform-provider-random/issues/66

When planning with this module, I did not observe the token to be printed in the console as described in that issue. But since there is now an AWS provider resource that is specifically intended to be treated as a secret ([`random_password`](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password)), it seems logical to use it here.

If you would accept a PR for this I can put something together.

rpdelaney:
## Description
- Use `random_password.result` instead of `random_id.hex` since this is a resource specifically intended for secrets

## Motivation and Context
Fixes #158

## Breaking Changes
None

## How Has This Been Tested?


aairey (approved):

aairey:
Not stale.

PR #262, opened 2022-03-14
fix: EFS encryption by default and variable support

nickmaccarthy:
## Description
* Enabled encryption on EFS by default.  
* Added `kms_key_id` support if a user wants to optionally encrypt their atlantis EFS with their own CMK
* Added `efs_` variables to turn these features on and off as necessary by a user
* Added `efs_additional_tags` so a user can add additional EFS tags as necessary, otherwise we will use the standard `local.tags` for this EFS module
* Added support for other portions of the `aws_efs_file_system` resource here, and set the appropriate variables so a user and turn things on and off as they wish.  https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/efs_file_system

## Motivation and Context

Not having EFS encrypted by default is a security risk and a violation in any mature organization.  Enabling this by default is best practice and will enable more and more users of this moudule to adopt EFS support for Atlantis

This [pr](https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/257) has been open, but doenst added kms_key support.  This PR addressed #257 and then some

## Breaking Changes

It should not break anything as the module stands today

## How Has This Been Tested?

I have not been able to test these changes since this PR has to come from my personal github account and I dont have access to an area to fully build out a new atlantis to test.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -428,6 +428,10 @@ resource "aws_route53_record" "atlantis" {
 resource "aws_efs_file_system" "this" {
   count = var.enable_ephemeral_storage ? 0 : 1
 
+  encrypted      = var.efs_encrption_enabled 
+  kms_key_id     = var.efs_kms_key_id
+  tags           = merge(var.efs_additional_tags, local.tags) 
+
   creation_token = var.name
 }
 
--- a/main.tf
+++ b/main.tf
@@ -424,15 +424,22 @@ resource "aws_route53_record" "atlantis" {
 ################################################################################
 # EFS
 ################################################################################
-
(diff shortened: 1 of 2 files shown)


pauloconnor:
The tags and being able to set what type of EFS fs would be really useful. I found the burstable EFS killed performance on my container.

ax-nmac:
Hello, what has to be done to get this PR approved and merged?  EFS is a very useful feature for us, and encryption is a must.  Everytime we update our Atlantis images we loose all of our plans and it causing confusion.  EFS is the way to go here, but the current implementation of EFS in this module is lacking, this PR addresses those issues.

Issue #266, opened 2022-03-29
Add support for ACM when using private hosted zones


michelmzs:
## Is your request related to a problem? Please describe.
The current module config does not support ACM certificates when a private hosted zone is in use, the aws_route53_zone data sets the private_zone filter to true, but the ACM cannot be used with this approach.

## Describe the solution you'd like.
Prepare the module to use the public hosted zone id to create the ACM validation record when the variable `route53_private_zone_name` is set. Otherwise, the module maintains with the current behavior.

## Describe alternatives you've considered.
1. Disable certificate generation
2. Creation of the ACM ins a separated module
3. Referenced the ACM arn.

PR #267, opened 2022-03-29
feat: Allow ACM certificate usage when using private hosted zones

michelmzs:
## Description

Allow ACM certificate usage when using private hosted zones

## Motivation and Context

Fixes [#266](https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/266)

## Breaking Changes

- Variables rename: `route53_zone_name` to `route53_public_zone_name`
- New variable: `route53_private_zone_name`
- Variable removal: `route53_private_zone`

## How Has This Been Tested?

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -206,7 +206,7 @@ variable "certificate_arn" {
 }
 
 variable "acm_certificate_domain_name" {
-  description = "Route53 domain name to use for ACM certificate. Route53 zone for this domain should be created in advance. Specify if it is different from value in `route53_zone_name`"
+  description = "Route53 domain name to use for ACM certificate. Route53 zone for this domain should be created in advance. Specify if it is different from value in `route53_public_zone_name`"
   type        = string
   default     = ""
 }
--- a/variables.tf
+++ b/variables.tf
@@ -212,8 +212,14 @@ variable "acm_certificate_domain_name" {
 }
 
 # Route53
-variable "route53_zone_name" {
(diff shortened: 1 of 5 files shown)


variables.tf
```diff
@@ -206,14 +206,20 @@ variable "certificate_arn" {
...
-  description = "Route53 domain name to use for ACM certificate. Route53 zone for this domain should be created in advance. Specify if it is different from value in `route53_zone_name`"
+  description = "Route53 domain name to use for ACM certificate. Route53 zone for this domain should be created in advance. Specify if it is different from value in `route53_public_zone_name`"
   type        = string
   default     = ""
 }
 
 # Route53
-variable "route53_zone_name" {
-  description = "Route53 zone name to create ACM certificate in and main A-record, without trailing dot"
+variable "route53_public_zone_name" {
```
antonbabenko, line 215: We can't simply rename existing variables because there are users who are using them.

Instead, let's keep the same variable names as before but add a new boolean/string variable(s) that will be responsible for what you are trying to solve in this PR.

michelmzs: I suggested this refactor because the current variable (`route53_private_zone: bool`) for private_zone usage are partially implemented, if the user just enable it, the ACM validation hangs until the Terraform timeouts (You cannot validate a ACM for private hosted zones), to proceed, the user needs to disable certificate ACM generation.
This refactor aims to fix this by creating a single flow for private hosted zone usage, I will revert the variable renaming.

PR #275, opened 2022-04-23
feat: Introduce IPv6 CIDR specific allow ALBs variables

velimir:
### Description

Allow users to provide IPV6 CIDR blocks for the AWS ALBs

### Motivation and Context

Github started using IPv6 addresses.

If Atlantis is configured to allow GitHub IPs (via `alb_ingress_cidr_blocks` using "github" provider), terraform would fail with the following error:

``` 
Error: Error authorizing security group rule type ingress: InvalidParameterValue: CIDR block 2606:50c0::/32 is malformed
```

Addressed the stale comments on #217 (Fixed #217)

### Breaking Changes

No breaking changes.

### How Has This Been Tested?

Tested manually.


main.tf
```diff
@@ -329,9 +329,10 @@ module "alb_https_sg" {
 
   name        = "${var.name}-alb-https"
   vpc_id      = local.vpc_id
-  description = "Security group with HTTPS ports open for specific IPv4 CIDR block (or everybody), egress ports are all world open"
+  description = "Security group with HTTPS ports open for specific IPv4 and IPv6 CIDR block (or everybody), egress ports are all world open"
```
antonbabenko, line 332: Changing the description will trigger recreation of the security group resource if I am not mistaken, right? Thinking aloud, if there are any downsides to doing this? What do you think?

antonbabenko (approved):

antonbabenko:
Please fix the CI before we can merge it.

main.tf
velimir: Yeah, no, good catch. Thank you. I'm reverting the change of the decription. I don't think these changes are important enough to justrify recreation of the secruity groups.

antonbabenko:
This PR is included in version 3.15.0 :tada:

PR #276, opened 2022-04-27
feat: Added support for GitHub app

### LINKED ISSUE #271: Allow running atlantis with github app instead of a user token

Since we want to use atlantis in a GitHub organization, I want to have this environment variables and secrets in my atlantis container without any github user tokens.:

```terraform
    {
      name  = "ATLANTIS_GH_APP_ID"
      value = local.github_app_id
    },
    {
      "name" : "ATLANTIS_WRITE_GIT_CREDS",
      "value" : "true",
    }
  ]

  custom_environment_secrets = [
    {
      name      = "ATLANTIS_GH_APP_KEY"
      valueFrom = local.atlantis_github_app_key_ssm_parameter_name
    }
  ]
```
The module currently depends on the presence of `var.atlantis_github_user_token`, which is used as a conditional in the creation of some resources and that brings a problem:
- Without the atlantis_github_user_token, although there may be other secrets in the environment(webhook_secret. github app key secret etc.), the tasks can not get their value.
https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/fab764c06444a0e72deb6c3cf92bdb4fbc31ba4b/main.tf#L567

I worked around this issue by modifying the module by adding some custom variables.

## Describe the solution you'd like.

I would like to create the atlantis module with a simple parameter like `use_github_bot = true` which will take care of setting up all the necessary environment variables/parameter store secrets for me.

abalcobia:
## Description
Allow users to utilize the GitHub App instead of a personal access token to execute Atlantis.

## Motivation and Context
Enable the use of a GitHub App with Atlantis. In this PR, four variables were added to the module and their dependencies: 
- `var.atlantis_github_app_key` 
- `var.atlantis_github_app_key_ssm_parameter_name`
- `var.atlantis_write_git_creds`
- `var.atlantis_github_app_id`

Note: It is not necessary to create a specific variable for the webhook secret in the GitHub App, the existing one can be used.

- fixes #271

ref:https://www.runatlantis.io/docs/access-credentials.html#generating-an-access-token (GitHub App section)

## Breaking Changes
None

## How Has This Been Tested?

Initially, a github app was created and consequently a private key and a secret webhook were associated with 
it. Posteriorly, a test repo was used with the inputs (atlantis_github_app_id, atlantis_github_app_key & atlantis_github_webhook_secret) from the github app to guarantee the communication between Atlantis and the GitHub app at the level of webhooks and comments and it was possible to guarantee this communication with success as well as the provisioning of resources.

### DIFF
--- a/examples/github-complete/README.md
+++ b/examples/github-complete/README.md
@@ -81,6 +81,7 @@ Go to https://eu-west-1.console.aws.amazon.com/ecs/home?region=eu-west-1#/settin
 | <a name="input_github_app_key"></a> [github\_app\_key](#input\_github\_app\_key) | The PEM encoded private key for the GitHub App | `string` | n/a | yes |
 | <a name="input_github_owner"></a> [github\_owner](#input\_github\_owner) | Github owner | `string` | n/a | yes |
 | <a name="input_github_repo_names"></a> [github\_repo\_names](#input\_github\_repo\_names) | List of Github repositories that should be monitored by Atlantis | `list(string)` | n/a | yes |
+| <a name="input_github_token"></a> [github\_token](#input\_github\_token) | Github token | `string` | n/a | yes |
 | <a name="input_github_user"></a> [github\_user](#input\_github\_user) | Github user for Atlantis to utilize when performing Github activities | `string` | n/a | yes |
 | <a name="input_github_webhook_secret"></a> [github\_webhook\_secret](#input\_github\_webhook\_secret) | Webhook secret | `string` | n/a | yes |
 
--- a/examples/github-complete/README.md
+++ b/examples/github-complete/README.md
@@ -1,10 +1,25 @@
-# Complete Atlantis example with GitHub Webhooks
+# Complete Atlantis example with GitHub App and Webhooks
 
 Configuration in this directory creates the necessary infrastructure and resources for running Atlantis on Fargate plus GitHub repository webhooks configured to Atlantis URL.
 
 An existing Route53 hosted zone and domain is required to deploy this example.
 
-GitHub's personal access token can be generated at https://github.com/settings/tokens
+The GitHub App can be generated using multiple methods:
+
+- You can follow Atlantis instructions depicted [here](https://www.runatlantis.io/docs/access-credentials.html#github-app). The Atlantis method mostly automates the GitHub App generation using [GitHub App Manifest](https://docs.github.com/en/developers/apps/building-github-apps/creating-a-github-app-from-a-manifest), but you need an exposed endpoint to complete the process.
+- The other method is to manually create the GitHub App as instructed [here](https://docs.github.com/en/developers/apps/building-github-apps/creating-a-github-app).
+
+Once you have your GitHub App registered you will be able to access/manage the required parameters:
+
+- `atlantis_github_app_id` to identify the GitHub app.
+- `atlantis_github_app_key` to interact with GitHub.
+- `atlantis_github_webhook_secret` to receive and validate incoming webhook invocations from GitHub.
+
+## GitHub Personal Access Token (PAT) is no longer recommended
+
+While still supported, the use of GitHub Personal Access Token (PAT) is no longer the recommended method in favor of GitHub App.
+
+[GitHub Apps](https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps) provide more control over repository access/permissions and does not require the use of bot accounts.
 
 ## Usage
 
@@ -62,10 +77,12 @@ Go to https://eu-west-1.console.aws.amazon.com/ecs/home?region=eu-west-1#/settin
 |------|-------------|------|---------|:--------:|
 | <a name="input_alb_ingress_cidr_blocks"></a> [alb\_ingress\_cidr\_blocks](#input\_alb\_ingress\_cidr\_blocks) | List of IPv4 CIDR ranges to use on all ingress rules of the ALB - use your personal IP in the form of `x.x.x.x/32` for restricted testing | `list(string)` | n/a | yes |
 | <a name="input_domain"></a> [domain](#input\_domain) | Route53 domain name to use for ACM certificate. Route53 zone for this domain should be created in advance | `string` | n/a | yes |
+| <a name="input_github_app_id"></a> [github\_app\_id](#input\_github\_app\_id) | GitHub App ID that is running the Atlantis command | `string` | n/a | yes |
+| <a name="input_github_app_key"></a> [github\_app\_key](#input\_github\_app\_key) | The PEM encoded private key for the GitHub App | `string` | n/a | yes |
 | <a name="input_github_owner"></a> [github\_owner](#input\_github\_owner) | Github owner | `string` | n/a | yes |
 | <a name="input_github_repo_names"></a> [github\_repo\_names](#input\_github\_repo\_names) | List of Github repositories that should be monitored by Atlantis | `list(string)` | n/a | yes |
-| <a name="input_github_token"></a> [github\_token](#input\_github\_token) | Github token | `string` | n/a | yes |
 | <a name="input_github_user"></a> [github\_user](#input\_github\_user) | Github user for Atlantis to utilize when performing Github activities | `string` | n/a | yes |
+| <a name="input_github_webhook_secret"></a> [github\_webhook\_secret](#input\_github\_webhook\_secret) | Webhook secret | `string` | n/a | yes |
 
 ## Outputs
 
--- a/examples/github-complete/README.md
+++ b/examples/github-complete/README.md
@@ -23,7 +23,7 @@ While still supported, the use of GitHub Personal Access Token (PAT) is no longe
 
 ## Usage
 
-To run this code you need to copy `terraform.tfvars.sample` into `terraform.tfvars` and update the values locally or specify them using environment variables (`TF_VAR_github_token=xxx`, `TF_VAR_github_owner=xxx`, etc.). Once ready, execute:
+To run this code you need to copy `terraform.tfvars.sample` into `terraform.tfvars` and update the values locally or specify them using environment variables (`TF_VAR_github_app_id=xxx`, `TF_VAR_github_owner=xxx`, etc.). Once ready, execute:
 
 ```bash
 $ terraform init
@@ -60,7 +60,6 @@ Go to https://eu-west-1.console.aws.amazon.com/ecs/home?region=eu-west-1#/settin
 |------|--------|---------|
 | <a name="module_atlantis"></a> [atlantis](#module\_atlantis) | ../../ | n/a |
 | <a name="module_atlantis_access_log_bucket"></a> [atlantis\_access\_log\_bucket](#module\_atlantis\_access\_log\_bucket) | terraform-aws-modules/s3-bucket/aws | ~> 3.0 |
-| <a name="module_github_repository_webhook"></a> [github\_repository\_webhook](#module\_github\_repository\_webhook) | ../../modules/github-repository-webhook | n/a |
 
 ## Resources
 
@@ -81,7 +80,6 @@ Go to https://eu-west-1.console.aws.amazon.com/ecs/home?region=eu-west-1#/settin
 | <a name="input_github_app_key"></a> [github\_app\_key](#input\_github\_app\_key) | The PEM encoded private key for the GitHub App | `string` | n/a | yes |
 | <a name="input_github_owner"></a> [github\_owner](#input\_github\_owner) | Github owner | `string` | n/a | yes |
 | <a name="input_github_repo_names"></a> [github\_repo\_names](#input\_github\_repo\_names) | List of Github repositories that should be monitored by Atlantis | `list(string)` | n/a | yes |
-| <a name="input_github_token"></a> [github\_token](#input\_github\_token) | Github token | `string` | n/a | yes |
 | <a name="input_github_user"></a> [github\_user](#input\_github\_user) | Github user for Atlantis to utilize when performing Github activities | `string` | n/a | yes |
 | <a name="input_github_webhook_secret"></a> [github\_webhook\_secret](#input\_github\_webhook\_secret) | Webhook secret | `string` | n/a | yes |
 
@@ -92,7 +90,5 @@ Go to https://eu-west-1.console.aws.amazon.com/ecs/home?region=eu-west-1#/settin
 | <a name="output_atlantis_repo_allowlist"></a> [atlantis\_repo\_allowlist](#output\_atlantis\_repo\_allowlist) | Git repositories where webhook should be created |
 | <a name="output_atlantis_url"></a> [atlantis\_url](#output\_atlantis\_url) | URL of Atlantis |
 | <a name="output_ecs_task_definition"></a> [ecs\_task\_definition](#output\_ecs\_task\_definition) | Task definition for ECS service (used for external triggers) |
-| <a name="output_github_webhook_secret"></a> [github\_webhook\_secret](#output\_github\_webhook\_secret) | Github webhook secret |
-| <a name="output_github_webhook_urls"></a> [github\_webhook\_urls](#output\_github\_webhook\_urls) | Github webhook URL |
 | <a name="output_task_role_arn"></a> [task\_role\_arn](#output\_task\_role\_arn) | The Atlantis ECS task role arn |
 <!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
--- a/examples/github-complete/README.md
+++ b/examples/github-complete/README.md
@@ -6,7 +6,7 @@ An existing Route53 hosted zone and domain is required to deploy this example.
 
 ## Usage
 
-To run this code you need to copy `terraform.tfvars.sample` into `terraform.tfvars` and update the values locally or specify them using environment variables (`TF_VAR_github_app_id=xxx`, `TF_VAR_github_owner=xxx`, etc.). Ensure that bootstrap_GitHub_app is `true`.  Once ready, execute:
+To run this code you need to copy `terraform.tfvars.sample` into `terraform.tfvars` and update the values locally or specify them using environment variables (`TF_VAR_github_app_id=xxx`, `TF_VAR_github_owner=xxx`, etc.). Ensure that `bootstrap_github_app` is `true`.  Once ready, execute:
 
 ```bash
 $ terraform init
--- a/examples/github-complete/main.tf
+++ b/examples/github-complete/main.tf
@@ -82,10 +82,24 @@ module "atlantis" {
   permissions_boundary = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:policy/cloud/developer-boundary-policy"
   path                 = "/delegatedadmin/developer/"
 
-  # Atlantis
-  atlantis_github_user       = var.github_user
-  atlantis_github_user_token = var.github_token
-  atlantis_repo_allowlist    = [for repo in var.github_repo_names : "github.com/${var.github_owner}/${repo}"]
+  # Atlantis w/ GitHub user
+
+  atlantis_github_user    = var.github_user
+  atlantis_repo_allowlist = [for repo in var.github_repo_names : "github.com/${var.github_owner}/${repo}"]
+
+  # Atlantis w/ GitHub app
+
+  ################################################################################
+  # Suggestion: instead of allocating the values of the atlantis_github_app_key
+  # and atlantis_github_webhook_secret in the tfvars file,it is suggested to
+  # upload the values in the AWS Parameter Store of the atlantis account and
+  # call the values via the data source function
+  # (e.g. data.aws_ssm_parameter.ghapp_key.value) for security reasons.
+  ################################################################################
+
+  atlantis_github_app_id         = var.github_app_id
+  atlantis_github_app_key        = var.github_app_key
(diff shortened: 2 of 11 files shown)


main.tf
```diff
@@ -14,15 +14,15 @@ locals {
   )}"
   atlantis_url_events = "${local.atlantis_url}/events"
 
-  # Include only one group of secrets - for github, gitlab or bitbucket
-  has_secrets = var.atlantis_gitlab_user_token != "" || var.atlantis_github_user_token != "" || var.atlantis_bitbucket_user_token != ""
+  # Include only one group of secrets - for github, github app,  gitlab or bitbucket
+  has_secrets = var.atlantis_gitlab_user_token != "" || var.atlantis_github_user_token != "" || var.atlantis_github_app_key != "" || var.atlantis_bitbucket_user_token != ""
```
bryantbiggs, line 18: I believe we can wrap all these in this (untested):
```suggestion
  has_secrets = try(coalesce(var.atlantis_gitlab_user_token, var.atlantis_github_user_token, var.atlantis_github_app_key, var.atlantis_bitbucket_user_token) != "", false)
```

bryantbiggs: or perhaps better yet:

```suggestion
  has_secrets = compact([var.atlantis_gitlab_user_token, var.atlantis_github_user_token, var.atlantis_github_app_key, var.atlantis_bitbucket_user_token]) > 0
```

main.tf
```diff
@@ -566,7 +584,7 @@ data "aws_iam_policy_document" "ecs_task_access_secrets_with_kms" {
 }
 
 resource "aws_iam_role_policy" "ecs_task_access_secrets" {
-  count = var.atlantis_github_user_token != "" || var.atlantis_gitlab_user_token != "" || var.atlantis_bitbucket_user_token != "" ? 1 : 0
+  count = var.atlantis_github_user_token != "" || var.atlantis_github_app_key != "" || var.atlantis_gitlab_user_token != "" || var.atlantis_bitbucket_user_token != "" ? 1 : 0
```
bryantbiggs, line 587:
```suggestion
  count = local.has_secrets ? 1 : 0
```

main.tf
```diff
@@ -14,15 +14,15 @@ locals {
...
 
-  # Include only one group of secrets - for github, gitlab or bitbucket
-  has_secrets = var.atlantis_gitlab_user_token != "" || var.atlantis_github_user_token != "" || var.atlantis_bitbucket_user_token != ""
+  # Include only one group of secrets - for github, github app,  gitlab or bitbucket
+  has_secrets = var.atlantis_gitlab_user_token != "" || var.atlantis_github_user_token != "" || var.atlantis_github_app_key != "" || var.atlantis_bitbucket_user_token != ""
 
-  # token
-  secret_name_key        = local.has_secrets ? var.atlantis_gitlab_user_token != "" ? "ATLANTIS_GITLAB_TOKEN" : var.atlantis_github_user_token != "" ? "ATLANTIS_GH_TOKEN" : "ATLANTIS_BITBUCKET_TOKEN" : ""
-  secret_name_value_from = local.has_secrets ? var.atlantis_gitlab_user_token != "" ? var.atlantis_gitlab_user_token_ssm_parameter_name : var.atlantis_github_user_token != "" ? var.atlantis_github_user_token_ssm_parameter_name : var.atlantis_bitbucket_user_token_ssm_parameter_name : ""
+  # token/key
```
bryantbiggs, line 20: oof, I'll have to check this out locally to see if I can follow or make sense of this. its getting a little too unwieldy

bryantbiggs:
Thanks for the PR @abalcobia  - since GitHub apps are more secure and the recommended way to provide access, lets update the `github-complete` to use this functionality and provide any docs that users would need to get this setup to use or validate

main.tf
abalcobia: I tested both suggestions, but only the first one went well. I made a commit with the proposed change, ran it locally and it worked fine with my test repo.

main.tf
abalcobia: I tested this suggestion and it ran fine. The change is included in the "code improvements" commit.

abalcobia:
Regarding the updates and the docs on examples, I will work on that in the next days.

calexandre:
@bryantbiggs  @antonbabenko I've helped @abalcobia with the latest changes. Can you guys review please?

examples/github-complete/main.tf
```diff
@@ -73,10 +73,26 @@ module "atlantis" {
   # Trusted roles
   trusted_principals = ["ssm.amazonaws.com"]
 
-  # Atlantis
+  # Atlantis w/ GitHub user
+
   atlantis_github_user       = var.github_user
```
bryantbiggs, line 78: Lets remove this - I don't think it would be correct to have both user based auth and app based

examples/github-complete/terraform.tfvars.sample
```diff
@@ -3,3 +3,6 @@ alb_ingress_cidr_blocks = ["x.x.x.x/32"]
 github_owner = "myorg"
 github_user = "atlantis"
```
bryantbiggs, line 4: user and token would be removed here as well

examples/github-complete/variables.tf
```diff
@@ -22,3 +22,18 @@ variable "github_user" {
   description = "Github user for Atlantis to utilize when performing Github activities"
   type        = string
 }
+
```
bryantbiggs, line 25: and user and token variables here

calexandre:
Thanks for the review @bryantbiggs! we'll make the requested changes asap 👍

abalcobia:
Hello @bryantbiggs! Just a question to clarify the approach, is it only supposed to remove the user and token variables, and respective dependencies, in the folder examples or throughout the entire repo?

bryantbiggs:
@abalcobia - correct, just remove those from the example. It doesn't make a lot of sense to have two authentication routes for one example, I don't even know if this works, so lets just show the "preferred" way which is to use the GitHub app authentication route in the example. Does that make sense? The user/token authentication route will still exist in the module itself, we're just changing the example to show using the GitHub app route.

abalcobia:
@bryantbiggs - yes, makes sense, thanks. We get the idea and throughout the day we will update the README.md file only in the examples folder.

calexandre:
@bryantbiggs  we've included instructions on how to generate a github app. We've also added a note stating that using GitHub PATs is no longer recommended (despite being still supported).

Let us know if you need any more changes.

calexandre:
@bryantbiggs we've updated the PR with the latest changes from the `master`.

Can we get this merged on the master, or is something missing?

abalcobia:
Hello @bryantbiggs @antonbabenko 

If you feel that any more changes need to be made please let us know. 

Thanks.

bryantbiggs:
I will test this today

calexandre:
@bryantbiggs  sorry to bother, but do you have any updates on this? We've updated the branch to the latest commit on master...

matthiassb:
Commenting to keep open

calexandre:
rebased and squashed the commit history

bryantbiggs:
apologies for the delay - just getting back around to this. I'm a bit confused: what is the benefit here for GitHub app support if GitHub PAT is still required?

calexandre:
Well, it depends if you want to keep supporting PAT based installations...
Do you want us to remove all PAT references?

bryantbiggs:
I don't think thats the way to look at it - there are two avenues:
1. User elects to use GitHub PAT (and only PAT) 
2. User elects to use GitHub App (and only App - no PAT)

I believe both can be supported for now to avoid a breaking change, but this current implementation does not satisfy point 2 above which makes the current implementation a bit confusing. Why use a GitHub app if a PAT is still required?

calexandre:
> I don't think thats the way to look at it - there are two avenues:
> 
> 1. User elects to use GitHub PAT (and only PAT)
> 2. User elects to use GitHub App (and only App - no PAT)
> 
> I believe both can be supported for now to avoid a breaking change, but this current implementation does not satisfy point 2 above which makes the current implementation a bit confusing. Why use a GitHub app if a PAT is still required?

I'm confused...All PAT related variables are optional..
Can you please request a change on where the current implementation does not satisfy the requirements? Maybe it is easier to understand what you mean...

examples/github-complete/terraform.tfvars.sample
bryantbiggs: @calexandre

bryantbiggs:
@calexandre simply try deploying the `github-complete` example, it does not work as defined in this PR currently. You can see why it won't work because a PAT is still required for the webhook https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/master/modules/github-repository-webhook/main.tf#L2-L4

calexandre:
Oh, that must have slipped on a previous rebase...
I will look into that in a moment!

Thank you for pointing it out 👍

calexandre:
Just note that it is not possible to configure the github provider using both approaches, you either need to choose the PAT of the APP approach, or use two providers...

bryantbiggs:
> Just note that it is not possible to configure the github provider using both approaches, you either need to choose the PAT of the APP approach, or use two providers...

Is it possible to set the variables to `null` as default - I was going to try this but the example wasn't working to start so didn't get to it. Might have to be part of a larger refactor and externalize the provider like usual

calexandre:
@bryantbiggs I re-added the `github_token` variable on all modules that require `github-repository-webhook` to make this work for both scenarios.

Please note that only Atlantis should benefit from the github app auth method as it is more suited for this type applications.

It is debatable whether the `github-repository-webhook` should use PAT or APP authentication; regardless of which method is preferred I believe it is outside of the scope of this PR, since it would have to be a different APP with a different set of permissions.

Note that Atlantis default/recommended application permissions don't involve any kind of repository settings interaction, which need to be changed in order to make it work to create webhooks.

calexandre:
> > Just note that it is not possible to configure the github provider using both approaches, you either need to choose the PAT of the APP approach, or use two providers...
> 
> Is it possible to set the variables to `null` as default - I was going to try this but the example wasn't working to start so didn't get to it. Might have to be part of a larger refactor and externalize the provider like usual

I was not able to setup the github provider to work with both methods in the same provider. The only way is to setup a second provider alias and use the APP auth on that one.
You are right, the correct way would be to take the provider out of the module so that the invoker controls the provider authentication.

examples/github-complete/terraform.tfvars.sample
calexandre: I cannot remove this because it would break the repository webhook creation.

examples/github-complete/variables.tf
calexandre: I cannot remove this because it would break the repository webhook creation.

calexandre:
@bryantbiggs  I also forgot to mention that with a GitHub App you no longer need to create a repository webhook for sending events... That part is configured on the GitHub App settings.

One suggestion to make the example a bit more elegant could be to only invoke the webhook module when the app authentication variables are not set...

bryantbiggs:
if thats the case then lets remove the webhook from the example. the webhook is only there to send the payload to Atlantis when there are changes

bryantbiggs:
(my bad - wrong button 🤣 )

calexandre:
> if thats the case then lets remove the webhook from the example. the webhook is only there to send the payload to Atlantis when there are changes

In that case I will remove all variables and references to the `github-repository-webhook` module (including the module itself).
Is that okay?

calexandre:
> > if thats the case then lets remove the webhook from the example. the webhook is only there to send the payload to Atlantis when there are changes
> 
> In that case I will remove all variables and references to the `github-repository-webhook` module (including the module itself). Is that okay?

@bryantbiggs could you please give more details on what you mean by removing the webhook from the example? You meant to remove all the webhook implementations, or just from the particular `github-complete` example while keeping the module, etc..?

On the other hand, we could also copy the existing example and name it to something like `github-complete-app-auth` and remove the webhook references from there. That way you keep both examples and preserve retrocompatibility.

Just let me know how you want to proceed so that we can get this merged :)

bryantbiggs:
Just remove the webhook implementation from the `github-complete` example. We'll leave the sub-module in the project so that we don't break the existing implementation for users

calexandre:
@bryantbiggs I've updated the `github-complete` example and removed the `github-repository-webhook` dependency from the example.

calexandre:
@bryantbiggs  is there anything else missing for merging this PR?

andyshinn:
Can we reopen this and get it merged?

calexandre:
I've rebased the branch so that you can review.
@antonbabenko @dynamike  let me know if you want to make any additional changes.

examples/github-complete/README.md
```diff
@@ -1,14 +1,29 @@
...
+# Complete Atlantis example with GitHub App and Webhooks
 
 Configuration in this directory creates the necessary infrastructure and resources for running Atlantis on Fargate plus GitHub repository webhooks configured to Atlantis URL.
 
 An existing Route53 hosted zone and domain is required to deploy this example.
 
-GitHub's personal access token can be generated at https://github.com/settings/tokens
+The GitHub App can be generated using multiple methods:
+
+- You can follow Atlantis instructions depicted [here](https://www.runatlantis.io/docs/access-credentials.html#github-app). The Atlantis method mostly automates the GitHub App generation using [GitHub App Manifest](https://docs.github.com/en/developers/apps/building-github-apps/creating-a-github-app-from-a-manifest), but you need an exposed endpoint to complete the process.
```
dynamike, line 9: This option doesn't seem possible because as you've mentioned you need an exposed endpoint to be able to setup the GitHub App. How would one go about doing that? Should I - 
 * Spin up a version of the `github-complete` example without a valid `github_app_id`, `github_app_key`, and `github_webhook_secret`
 * Register the GitHub app via the exposed endpoint
 * Re-run terraform with the GitHub app values?

More detail here is needed if this is the preferred path.

examples/github-complete/README.md
```diff
@@ -1,14 +1,29 @@
...
+- The other method is to manually create the GitHub App as instructed [here](https://docs.github.com/en/developers/apps/building-github-apps/creating-a-github-app).
```
dynamike, line 10: I also ran into a wall with this method. The Github docs on setting up a Github App  are very generic and doesn't explain how I should put values in for sections like `Identifying and authorizing users` or `Webhook URL` to work with Atlantis. So I'm left without a path forward to test this out.

examples/github-complete/README.md
```diff
@@ -1,14 +1,29 @@
...
+
+Once you have your GitHub App registered you will be able to access/manage the required parameters:
+
+- `atlantis_github_app_id` to identify the GitHub app.
+- `atlantis_github_app_key` to interact with GitHub.
+- `atlantis_github_webhook_secret` to receive and validate incoming webhook invocations from GitHub.
+
+## GitHub Personal Access Token (PAT) is no longer recommended
+
+While still supported, the use of GitHub Personal Access Token (PAT) is no longer the recommended method in favor of GitHub App.
```
dynamike, line 20: I think it makes sense to support both the Github App **and** the Personal Access Token (PAT) method in this module. While it's true you get fine grained access with Github Apps, there is a GitHub beta feature to allow you to created [PATs with fine grained access](https://github.blog/2022-10-18-introducing-fine-grained-personal-access-tokens-for-github/). So the benefit of a Github App vs a PAT become less clear.

dynamike (changes requested):
I wasn't really able to test this feature out because of a difficulty in bootstrapping GitHub apps, which I describe in this review. Let me know what you think as I think this would be a handy feature.

examples/github-complete/README.md
calexandre: Note that despite the PAT's new features, you still require a valid user to create the PAT for. This means consuming an additional license for certain organizations and using an authentication mechanism whose purpose is for users and not apps/services.

Apart from fine grained access, there are multiple advantages by using a GitHub App, one of them being the extensive webhook configuration options, among other things.

Since Atlantis is an app, it fits the GitHub App's purpose.

examples/github-complete/README.md
calexandre: I understand your pain... it's not very easy to bootstrap Atlantis using the official documentation.

What the Atlantis team did, was to try to ease the process of setting up the GitHub App by using the GitHub App manifest configuration method - this 'manifest' method, involves the Atlantis runtime to provide all the requirements for GitHub to create the app on Atlantis behalf. Of course, this raises security concerns: 
- GitHub needs to trust the exposed endpoint (this means not only being exposed, but needs to have a valid SSL certificate)
- You need to expose your app to the internet and follow the 'one-time' setup procedure to get the required secrets `github_app_id`, `github_app_key`, and `github_webhook_secret`.
- This method not only provides you the secrets, but also configures the GitHub App with the required permissions as depicted by the Atlantis team.

**TL;DR:** The steps you mentioned are correct. That's what you need to do in order to setup the GitHub App according to Atlantis official documentation.

examples/github-complete/README.md
calexandre: I understand the confusion. This is an advanced setup which requires the user to be comfortable with GitHub Apps, hence the above recommended path depicted by the Atlantis team.

Let me try and make this clearer:

- You create a GitHub App and give it a name - that name must be unique across the world (you can change it later).
- Provide a valid Homepage URL (this can be the atlantis server url, for instance `https://atlantis.mydomain.com`)
- Provide a valid Webhook URL. The Atlantis webhook server path is located by default at `https://atlantis.mydomain.com/events`
- Generate a Webhook Secret - this is used for Atlantis to trust the deliveries. This is your `github_webhook_secret`.
- Generate a Private Key - this is your `github_app_key`
- On the App's settings page (at the top) you find the App ID. This is your `github_app_id`
- On the **Permissions & Events** you need to setup all the permissions and events according to [Atlantis documentation](https://www.runatlantis.io/docs/access-credentials.html#permissions)

Now you need to Install the App on your organization.

A self-provisioned GitHub App usually has two parts: the App and the Installation.
- The App part is the first step and its where you setup all the requirements, such as authentication, webhook, permissions, etc...
- The Installation part is where you add the created App to an organization/personal-account. It is on the installation page where you setup which repositories the application can access and receive events from.

examples/github-complete/README.md
calexandre: I'd like to add that if your organization has an SSO, using a PAT is a bit cumbersome.

examples/github-complete/README.md
bryantbiggs: All of this is great info to add to the module docs for users

dynamike (commented):
I was able to get this working using the method in the Atlantis documentation, but I had to make some changes to the github-complete example for it to work correctly. You can see the changes I made over at https://github.com/nosportugal/terraform-aws-atlantis/compare/github-app-support...dynamike:terraform-aws-atlantis:github-app-support?expand=1 as I don't really have a way to incorporate those changes in this PR.

calexandre:
> I was able to get this working using the method in the Atlantis documentation, but I had to make some changes to the github-complete example for it to work correctly. You can see the changes I made over at https://github.com/nosportugal/terraform-aws-atlantis/compare/github-app-support...dynamike:terraform-aws-atlantis:github-app-support?expand=1 as I don't really have a way to incorporate those changes in this PR.

Looks great! Just waiting for the review on the branch...

dynamike (approved):
@antonbabenko I think this is ready to go. I've confirmed there aren't breaking changes which means using a traditional webhook or a Github App both work. The github-complete example is also functional using the Github App method. 

The one thing that I would note is that [ATLANTIS_WRITE_GIT_CREDS](https://www.runatlantis.io/docs/server-configuration.html#write-git-creds) is now defaulted to true because it is required if you are using the GitHub App method, but this will also enable the flag for folks not using the Github App.

examples/github-complete/README.md
```diff
@@ -1,21 +1,27 @@
...
 Configuration in this directory creates the necessary infrastructure and resources for running Atlantis on Fargate plus GitHub repository webhooks configured to Atlantis URL.
 
 An existing Route53 hosted zone and domain is required to deploy this example.
 
-GitHub's personal access token can be generated at https://github.com/settings/tokens
-
 ## Usage
 
-To run this code you need to copy `terraform.tfvars.sample` into `terraform.tfvars` and update the values locally or specify them using environment variables (`TF_VAR_github_token=xxx`, `TF_VAR_github_owner=xxx`, etc.). Once ready, execute:
+To run this code you need to copy `terraform.tfvars.sample` into `terraform.tfvars` and update the values locally or specify them using environment variables (`TF_VAR_github_app_id=xxx`, `TF_VAR_github_owner=xxx`, etc.). Ensure that bootstrap_GitHub_app is `true`.  Once ready, execute:
```
antonbabenko, line 9:
```suggestion
To run this code you need to copy `terraform.tfvars.sample` into `terraform.tfvars` and update the values locally or specify them using environment variables (`TF_VAR_github_app_id=xxx`, `TF_VAR_github_owner=xxx`, etc.). Ensure that `bootstrap_github_app` is `true`.  Once ready, execute:
```

antonbabenko (approved):

dynamike:
@calexandre Fixes for the tflint warnings are in https://github.com/nosportugal/terraform-aws-atlantis/pull/7

dynamike:
@antonbabenko I think this one is ready to merge when you're able. Thanks!

antonbabenko:
This PR is included in version 3.23.0 :tada:

andyshinn:
Thanks for getting this over the finish line everyone!

abalcobia:
Thank you to everyone involved for the effort made to get the PR integrated!

Issue #272, opened 2022-04-06
If statement in for_each aws_efs_mount_target not working properly


sschrijver-pon:
## Description

When `var.enable_ephemeral_storage` is `true`, the `if` condition in `aws_efs_mount_target` isn't behaving as expected and is giving the following exception:

```
Error: Invalid index

  on main.tf line 438, in resource "aws_efs_mount_target" "this":
 438:   file_system_id  = aws_efs_file_system.this[0].id
    |----------------
    | aws_efs_file_system.this is empty tuple

The given key does not identify an element in this collection value.
```

Suggestion to change: 

```
resource "aws_efs_mount_target" "this" {
  # we coalescelist in order to specify the resource keys when we create the subnets using the VPC or they're specified for us.  This works around the for_each value depends on attributes which can't be determined until apply error
  for_each = {
    for k, v in zipmap(coalescelist(var.private_subnets, var.private_subnet_ids), local.private_subnet_ids) : k => v
    if var.enable_ephemeral_storage == false
  }
```
To:

```
resource "aws_efs_mount_target" "this" {
  # we coalescelist in order to specify the resource keys when we create the subnets using the VPC or they're specified for us.  This works around the for_each value depends on attributes which can't be determined until apply error
  for_each = {
    for k, v in zipmap(coalescelist(var.private_subnets, var.private_subnet_ids), local.private_subnet_ids) : k => v
    if !var.enable_ephemeral_storage
  }
```
## Versions

- Module version: 3.14.0

- Terraform version: 0.14.11
- Terragrunt version: 0.28.24
- Provider version(s):
```
+ provider registry.terraform.io/hashicorp/aws v3.70.0
+ provider registry.terraform.io/hashicorp/local v2.2.2
+ provider registry.terraform.io/hashicorp/random v3.1.2
```

## Reproduction Code

```
terraform {
  source = "git::git@github.com:terraform-aws-modules/terraform-aws-atlantis.git//?ref=tags/v3.14.0"

}

inputs = {
  ...
  enable_ephemeral_storage  = true
  ...
}
```
## Expected behavior

Skipping aws_efs_mount_target resources

## Actual behavior

Not skipping aws_efs_mount_target resources

PR #281, opened 2022-05-18
fix: EFS mount target for_each does not work if ephemeral storage is enabled

greggilbert:
## Description
This PR reworks the `aws_efs_mount_target` `for_each` to handle enable ephemeral storage better. This is a reworking of https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/272.

## Motivation and Context

Versions:
* Terraform v1.1.9
* terragrunt version v0.37.1
* AWS provider 4.14.0

I was setting `enable_ephemeral_storage = true`, but despite that, I was getting this error:

```
│ Error: Invalid for_each argument
│
│   on .terraform/modules/atlantis.atlantis/main.tf line 454, in resource "aws_efs_mount_target" "this":
│  454:   for_each = {
│  455:     for k, v in zipmap(coalescelist(var.private_subnets, var.private_subnet_ids), local.private_subnet_ids) : k => v
│  456:     if var.enable_ephemeral_storage == false
│  457:   }
│     ├────────────────
│     │ local.private_subnet_ids is list of string with 2 elements
│     │ var.enable_ephemeral_storage is true
│     │ var.private_subnet_ids is list of string with 2 elements
│     │ var.private_subnets is empty list of string
│
│ The "for_each" value depends on resource attributes that cannot be
│ determined until apply, so Terraform cannot predict how many instances will
│ be created. To work around this, use the -target argument to first apply
│ only the resources that the for_each depends on.
```

## Breaking Changes
This only breaks the latest version if you're enabling ephemeral storage.

## How Has This Been Tested?

I made the change locally and was able to `terraform plan` correctly.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -451,10 +451,11 @@ resource "aws_efs_file_system" "this" {
 
 resource "aws_efs_mount_target" "this" {
   # we coalescelist in order to specify the resource keys when we create the subnets using the VPC or they're specified for us.  This works around the for_each value depends on attributes which can't be determined until apply error
-  for_each = {
-    for k, v in zipmap(coalescelist(var.private_subnets, var.private_subnet_ids), local.private_subnet_ids) : k => v
-    if var.enable_ephemeral_storage == false
-  }
+  for_each = (
+    var.enable_ephemeral_storage == false
+    ? { for k, v in zipmap(coalescelist(var.private_subnets, var.private_subnet_ids), local.private_subnet_ids) : k => v }
+    : {}
+  )
 
   file_system_id  = aws_efs_file_system.this[0].id
   subnet_id       = each.value


main.tf
```diff
@@ -451,10 +451,11 @@ resource "aws_efs_file_system" "this" {
 
 resource "aws_efs_mount_target" "this" {
   # we coalescelist in order to specify the resource keys when we create the subnets using the VPC or they're specified for us.  This works around the for_each value depends on attributes which can't be determined until apply error
-  for_each = {
-    for k, v in zipmap(coalescelist(var.private_subnets, var.private_subnet_ids), local.private_subnet_ids) : k => v
-    if var.enable_ephemeral_storage == false
-  }
+  for_each = (
+    var.enable_ephemeral_storage == false
```
antonbabenko, line 455: I don't think this change to the code is really necessary because it is identical to the existing one which works a bit better.

I can't find an exact issue in [Terraform core](https://github.com/hashicorp/terraform) but it is related to `map` != `object` and we have this approach in all other modules already.

antonbabenko (commented):
@bryantbiggs WDYT about this issue? There were several reports about it since ephemeral EFS was added to the module.

main.tf
greggilbert: In its current form, it throws this error:
```
The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the resources that the for_each depends on.
```
So you can't apply the whole module, you need to somehow figure out which pieces should get applied first and apply those. Seems like a dealbreaker for using the module.

bryantbiggs: I believe its because we are using `zipmap()` which converts computed resources into map keys and violates https://github.com/hashicorp/terraform/issues/4149

I think it would be better if we used `count` instead of `for_each` - we don't have a way of creating a static key for the `for_each` maps which means we'll always suffer against https://github.com/hashicorp/terraform/issues/4149

so line 454 becomes:

```hcl
  count = var.enable_ephemeral_storage == false ? local.private_subnet_ids : []
```

greggilbert: It seems like you still need the `coalescelist`, though? Couldn't you do something like
```
for_each = (var.enable_ephemeral_storage == false ? toset(coalescelist(...)) : [])
```
Like that `efs_mount_target` resource isn't even using `each.key`, so it doesn't need to be a map?

bryantbiggs: `for_each` here only accepts maps as input

greggilbert: Ah, gotcha.

What's the best way to proceed here, then?

greggilbert: @bryantbiggs @antonbabenko Checking in to see if I can help get this through.

antonbabenko: @greggilbert I think the best way would be if you could reproduce the problem (you could), fix the code, so that it works well for your and other cases.

Maybe simplification is necessary as @bryantbiggs mentioned about `zipmap()` but I don't think it is related.

The most important is to make it work for your case and the opposite one.

Sorry, that I can't provide a better solution without diving into this issue myself.

PR #285, opened 2022-06-08
fix: Invalid for_each argument error with policies_arn

### LINKED ISSUE #284: Referencing a policy created within a module which implements this module throws invalid for_each argument error
## Description

If you create a policy and reference it in the `policies_arn` variable in this module, you receive an invalid for_each argument error.  You can find a reproduction of this in the [policy-apply-fail](https://github.com/MarkIannucci/terraform-aws-atlantis/tree/policy-apply-fail) branch in my fork.

I have a fix in my head that I will submit as a PR shortly.  

## Versions

- Module version [Required]:

- Terraform version:

- Provider version(s):

## Reproduction Code [Required]

Clone [my fork](https://github.com/MarkIannucci/terraform-aws-atlantis).
```bash
git switch -c policy-apply-fail
cd ./examples/github-complete
terraform plan
```
enter `["*"]` at the repos prompt.

## Expected behavior

No plan error

## Actual behavior

Terraform throws invalid for_each argument error.

### Terminal Output Screenshot(s)

image

MarkIannucci:
## Description
Converts the list of strings in `policies_arn` so the `for_each` which creates the role can iterate over parameterized input.

## Motivation and Context
Fixes #284 .

## Breaking Changes
I don't think so, though people who apply future versions of this module may notice a policy get destroyed and recreated because the index in the `for_each` changed.  I'm not 100% sure.

## How Has This Been Tested?
- [x] I cherrypicked the commit in this PR and put it in the [policy-apply-fail-fixed branch in my fork](https://github.com/MarkIannucci/terraform-aws-atlantis/tree/policy-apply-fail-fixed).  When I ran `terraform plan` it fixed the issue.


main.tf
```diff
@@ -527,7 +528,7 @@ resource "aws_iam_role" "ecs_task_execution" {
 }
 
 resource "aws_iam_role_policy_attachment" "ecs_task_execution" {
-  for_each = toset(local.policies_arn)
+  for_each = local.policies_arn_map
```
antonbabenko, line 531:
```suggestion
  for_each = { for k, v in local.policies_arn : k => v }
```

main.tf
```diff
@@ -106,7 +106,8 @@ locals {
     var.tags,
   )
 
-  policies_arn = var.policies_arn != null ? var.policies_arn : ["arn:${data.aws_partition.current.partition}:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"]
+  policies_arn     = var.policies_arn != null ? var.policies_arn : ["arn:${data.aws_partition.current.partition}:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"]
+  policies_arn_map = { for idx, arn in local.policies_arn : idx => arn }
```
antonbabenko, line 110:
```suggestion
```

main.tf
```diff
@@ -106,7 +106,8 @@ locals {
     var.tags,
   )
 
-  policies_arn = var.policies_arn != null ? var.policies_arn : ["arn:${data.aws_partition.current.partition}:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"]
+  policies_arn     = var.policies_arn != null ? var.policies_arn : ["arn:${data.aws_partition.current.partition}:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"]
```
antonbabenko, line 109:
```suggestion
  policies_arn = var.policies_arn != null ? var.policies_arn : ["arn:${data.aws_partition.current.partition}:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"]
```

main.tf
antonbabenko: `k` and `v` are more common in these modules (I believe).

antonbabenko (approved):
LGTM. This will trigger the recreation of policy attachments since the index is changed, but it is a valid fix that we have applied lately in other modules (IAM, EKS...).

@bryantbiggs WDYT?

bryantbiggs:
its a good change but we have to make a few more changes to make this work - specifically the `var.policies_arn` and `local.policies_arn` should probably be turned into maps where the keys are static strings. 

We could also consider splitting out the attachment of the `AmazonECSTaskExecutionRolePolicy` to its own attachment resource with a boolean variable like `attach_task_execution_policy` which defaults to `true`. And then the 2nd policy attachment resource would just `for_each = var.additional_policies` which is a map of `{ static_string = policy_arn }` which will support using passing policies generated through resources like `{ foo = aws_iam_role_policy.foo.arn }` which I think is the goal of this PR. This would remove [this logic](https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/1cb9812e21417828cfaa555d69a1cfc2a4a1ac46/main.tf#L109) that is not great (if someone wants to attach additional policies they also have to re-specify the task execution policy)

MarkIannucci:
I'm not familiar with when breaking changes are accepted or not.  @bryantbiggs, are we OK with a breaking change for this fix?  I think we are because your statement about converting `var.policies_arn` into a map would be one.  If we are, I'm happy to rework this PR to split out the `AmazonECSTaskExecutionRolePolicy` policy into a boolean variable and create the `additional_policies` variable.

I thought of a way to do this without a breaking change, but it would require a lot of logic to avoid a second attachment of the `AmazonECSTaskExecutionRolePolicy`, which I don't want to write and I don't think we want to maintain given our desire to remove the logic below:
https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/1cb9812e21417828cfaa555d69a1cfc2a4a1ac46/main.tf#L109

MarkIannucci:
I spent a bit more time thinking about this and I may have a solution which avoids the breaking change and implements the suggestion to create the `attach_task_execution_policy` boolean variable... I'll write an additional post with where I settle shortly.

MarkIannucci:
I agree, the data structure you propose above is ideal :)  Thank you for the review.

It was an interesting thought exercise for me to write the code to deal with this in a backwards compatible manner.  It produces the following data structure in pseudo-code:

```hcl
local.policies_arn_map = {
  "0" = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy" # assuming `attach_task_execution_policy` is true or the service-role is included in var.policies_arn
  "1" = other policies specified in var.policies_arn ...
}
```

This comes at the cost of this complexity:
https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/ad146f8ccf2fb695dae58f06054cfe4bb151a91e/main.tf#L109-L111

So we pick our poison.  
1. Declare a breaking change, remove `var.policies_arn` and create a `additional_policies_map` variable which we merge with the `attach_task_execution_policy` boolean variable to produce the local `role_policy_arns` map
2. Avoid a breaking change by accepting the complexity above.  This also has the flaw of a change in the order of the items listed will cause stuff to be destroyed replaced when not necessary.

MarkIannucci:
Sigh... I thought merging the default branch into this branch would restart the clock.  It did not.

matthiassb:
any updates here?

PR #286, opened 2022-06-10
feat: Allows looking up SGs by name & attach them to the ALB

titanlien:
## Description

Allows looking up SGs by name & attach them to the ALB

## How Has This Been Tested?


antonbabenko:
There is already `var.security_group_ids`, so you need to do data source lookup (using any filters you want - by tags, by name, etc) outside of the module and pass the value to this module.

PR #287, opened 2022-06-13
feat: Ensure EFS is owned by the posix_user to avoid permission issues

affelaffen:
## Description
When specifying a user for running the container, e.g. using the `100:1000` Atlantis user on the default image, the `aws_efs_access_point` is correctly configured with said `posix_user`, but the ownership of the root directory is left as the default `root:root`. This can result in different permission issues, one of which is described in https://github.com/runatlantis/atlantis/issues/2221. By specifying the `creation_info`, we ensure that the user provided as the executing posix user also owns the directory. If desired, this change can be put behind a feature variable.

## Motivation and Context
The issue described in https://github.com/runatlantis/atlantis/issues/2221 is the primary motivation for this change.

## Breaking Changes
No breaking changes as such, but this will force recreation of an existing aws_efs_access_point and thus the aws_ecs_task_definition.

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -469,6 +469,15 @@ resource "aws_efs_access_point" "this" {
     gid = local.gid
     uid = local.uid
   }
+
+  root_directory {
+    path = "/"
+    creation_info {
+      owner_gid   = local.gid
+      owner_uid   = local.uid
+      permissions = 0750
+    }
+  }
 }
 
 ################################################################################
--- a/main.tf
+++ b/main.tf
@@ -471,7 +471,7 @@ resource "aws_efs_access_point" "this" {
   }
 
   root_directory {
-    path = "/"
+    path = "/home/atlantis"
     creation_info {
       owner_gid   = local.gid
       owner_uid   = local.uid


jensenbox:
@affelaffen I applied this change locally and rebuilt my ECS cluster, EFS volume and EFS Access point and I am still encountering the `unsafe repository is owned by someone else` issue.

Any suggestion here?

affelaffen:
> @affelaffen I applied this change locally and rebuilt my ECS cluster, EFS volume and EFS Access point and I am still encountering the `unsafe repository is owned by someone else` issue.
> 
> Any suggestion here?

Do you run the module with `user = "100:1000"`?

jensenbox:
I did - with and without. Each time deleting everything in AWS. I can run
it all again tomorrow but I would end up in 1 of 3 possible situations:

1) The original issue - git complaining about an unsafe repo
2) The container would not start because of permission issues
3) The container would start but Atlantis could not run a plan because of
permission issues

I will have to run it again tomorrow and get more specifics.

The way I have worked around the issue is to use ephemeral container
storage - not ideal but was the only way to move forward.

affelaffen:
Since I am using a slightly modified entrypoint script, I did not experience this issue. I tried with a completely new setup and the default image and also ran into the issue with permissions within the `/home/atlantis` folder. I modified the `aws_efs_access_point` path to point  directly to the home folder instead, and this seems to have solved the issue.

userhas404d:
> Since I am using a slightly modified entrypoint script, I did not experience this issue. I tried with a completely new setup and the default image and also ran into the issue with permissions within the `/home/atlantis` folder. I modified the `aws_efs_access_point` path to point directly to the home folder instead, and this seems to have solved the issue.

@affelaffen confirmed this works - thanks!

antonbabenko:
This PR is included in version 3.17.0 :tada:

antonbabenko:
Thank you all for the fix and confirmation that it works!

pauloconnor:
This is a breaking change - if you have been running this previously, and have mounted `/` as your path, changing it to `/home/atlantis` will result in losing all your previous checkouts and plans. In order to fix it, you need to create a new `/home/atlantis` in your existing mount, then copy everything over, ensuring permissions are updated

PR #288, opened 2022-06-20
feat: Overwrite the git access token in SSM when rotate it

titanlien:
## Description

As title

## Motivation and Context

When we want to rotate git repository access token, we have to use `overwrite` attribute to overwrite it

## Breaking Changes

No

## How Has This Been Tested?


MotoMagellan:
From the perspective of my organization’s workflow, I consider this a bugfix as we have manually patched in a workaround to address our token rotation requirements.

titanlien:
> 

Thanks for support this, we should fix it in the right place. It could be feature or bug fix.

PR #279, opened 2022-05-15
feat: Add support for runtime_platform config block

zahorniak:
## Description
These changes add support for `runtime_platform` configuration block to `aws_ecs_task_definition`. This option allows running Atlantis on a custom Operating system and custom Architecture on Fargate.

## Motivation and Context
Closes #273 

## Breaking Changes
The minimum AWS provider version was updated 3.45 -> 3.69 ([AWS provider PR](https://github.com/hashicorp/terraform-provider-aws/pull/22016))

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

PR #291, opened 2022-06-22
fix: Set correct default for runtime_platform to ensure idempotence

affelaffen:
## Description
Changed the default value for the `runtime_platform` to avoid recreation of the `aws_ecs_task_definition` on every apply when not specifying any `runtime_platform` values. Since the dynamic block for the `runtime_platform` checks if the variable is null before trying to set a value and the current default value is set to an empty map, it will always try to add an empty map to the task definition. This causes terraform to consider it a change, which forces replacement of both the `aws_ecs_task_definition` and the `aws_ecs_service`.

## Motivation and Context
The issue was introduced in https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/279, which added this support for runtime_platform.

## Breaking Changes
No breaking changes.

## How Has This Been Tested?


antonbabenko:
This PR is included in version 3.17.1 :tada:

PR #299, opened 2022-07-06
fix: Unintentionally exposed endpoints when enabling github webhooks

casperbiering:
## Description
When following the guide to allow access for github webhooks, then you're actually incorrectly exposing endpoints in the following ways:
1) All endpoints is allowed from github IPs
2) `/events` endpoint is allowed from the entire world

## Motivation and Context
Limiting exposed endpoints is a good security practice.

## Breaking Changes
It shouldn't except:
1) abusing the rules to have github webhooks hit other endpoints than `/events`
2) abusing the `allow_github_webhooks` option to allow webhooks from other systems than github

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,6 @@ alb_authenticate_cognito = {
 If you are using one of the authentication methods above along with managed GitHub (not self-hosted enterprise version), you'll need to allow unauthenticated access to GitHub's Webhook static IPs:
 
 ```hcl
-allow_unauthenticated_access = true
 allow_github_webhooks        = true
 ```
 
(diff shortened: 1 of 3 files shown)


casperbiering:
This is still relevant

casperbiering:
This is still relevant

casperbiering:
@bryantbiggs @antonbabenko Can you guys review please?

dynamike:
I think the overall intent of this PR is valid, but doesn't appear to work the way I was expecting and after a bit of experimentation, I think I understand why. 

The default behavior of the ALB routing rules is to forward traffic **if** you don't specify an ALB authentication method. 
image

So any forwarding rules limiting `/events` to GitHub only IPs is ignored because by default all traffic is allowed. The only way I think this can be resolved is to have a fixed response (e.g., 401 Unauthorized) either as the default response or as a low priority listener rule.

PR #309, opened 2022-09-07
feat: Add output with public IP for whitelisting

yury-kustov-sb:
## Description
Output variable vpc_nat_public_ips  added 

## Motivation and Context
Atlantis can use internal resources like hashicorp vault that requires whitelisting so it will be very convinient for a lot of users to integrate the module into whole iaac

## How Has This Been Tested?
Locally tested


antonbabenko:
This PR is included in version 3.19.0 :tada:

PR #310, opened 2022-09-24
Role policy allowing ecs task role to assume another role

SamirMarin:
## Description

This PR Gives a `aws_iam_role_policy` to the ecs_task role allowing it to assume the role passed

## Motivation and Context

- with out this if using the assume role approach to run atlantis there seems to be no other way to give the ecs_task role the permissions to assume the role.

NA

## Breaking Changes

No, if variable not specified continues to work as is.

## How Has This Been Tested?

I've build an atlantis server based of the the [github-complete](https://github.com/terraform-aws-modules/terraform-aws-atlantis/tree/master/examples/github-complete) example, and ran this deployment off my fork. I didn't update the example here because the change is optional if not specified continues to work as before.

no breaking changes with the change

I have build an atlantis fargate server module, i've used the the assume role functionality, By adding 

```tf
 module "atlantis" {
    source = "github.com/SamirMarin/terraform-aws-atlantis?ref=option-to-allow-assume-role-policy"

    iam_assume_role_arn = "arn:aws:iam::${local.account_id}:role/ci-runner"

  .....
  }


```
after this is applied it gives the ecs task role the permissions to assume ci-runner, I'm able to run atlantis plan assuming the the role specified in the `iam_assume_role_arn`

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,7 @@ allow_github_webhooks        = true
 | [aws_efs_file_system.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/efs_file_system) | resource |
 | [aws_efs_mount_target.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/efs_mount_target) | resource |
 | [aws_iam_role.ecs_task_execution](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role) | resource |
+| [aws_iam_role_policy.ecs_task_access_assume](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role_policy) | resource |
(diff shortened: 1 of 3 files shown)


bryantbiggs:
this is not correct - regardless, we already have `var.trusted_entities` for assuming the role

SamirMarin:
> `trusted_entities`

@bryantbiggs 
is trusted policies not to allow other roles to assume the task runner role? I tried adding the role I want to assume there it did not work, I don't think that lets task runner assume other roles to run atlantis.


For reference I'm trying to add a permission policy like this https://github.com/runatlantis/atlantis/issues/856#issuecomment-674276478

I think trusted entitites just adds role trust policy, allowing other roles to assume task runner. I want task runner to be able to assume another role.

I'm not really seeing a way to do this here, unless I'm missing something. If I'm missing something iIf you point me in the right direction it be appreciated.

bryantbiggs:
@SamirMarin perhaps next time a better set of details and context would help clarify what you are after

This now sounds like you are having the Atlantis role assume a different role which is still supported by this module through `var.policies_arn`.

Issue #269, opened 2022-04-02
Cannot specify more than one extra_container_definitions


andyshinn:
## Description

When specifying more than one item in list of `extra_container_definitions` there is an error.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]: 3.13.1

- Terraform version:
```
Terraform v1.1.4
on darwin_amd64
+ provider registry.terraform.io/datadog/datadog v3.8.1
+ provider registry.terraform.io/hashicorp/aws v3.73.0
+ provider registry.terraform.io/hashicorp/local v2.2.2
+ provider registry.terraform.io/hashicorp/random v3.1.0
+ provider registry.terraform.io/taiidani/jenkins v0.9.0
```
- Provider version(s):

```
terraform providers -version
Terraform v1.1.4
on darwin_amd64
```

## Reproduction Code [Required]

```
module "container_definition_fluentbit" {
  source  = "cloudposse/ecs-container-definition/aws"
  version = "v0.58.1"

  container_name  = "log-router"
  container_image = "amazon/aws-for-fluent-bit:2.23.3"
  essential       = true

  firelens_configuration = {
    type = "fluentbit"

    options = {
      config-file-type        = "file"
      config-file-value       = "/fluent-bit/configs/parse-json.conf"
      enable-ecs-log-metadata = "true"
    }
  }
}

module "container_definition_datadog" {
  source  = "cloudposse/ecs-container-definition/aws"
  version = "v0.58.1"

  container_name  = "datadog-agent"
  container_image = local.datadog_image
  essential       = true
  environment = [
    {
      name  = "DD_API_KEY"
      value = "fake"
    },
    {
      name  = "ECS_FARGATE"
      value = "true"
    },
    {
      name  = "DD_APM_ENABLED"
      value = "true"
    },
    {
      name  = "DD_SITE"
      value = "ddog-gov.com"
    },
    {
      name  = "DD_LOGS_ENABLED"
      value = "true"
    }
  ]

  firelens_configuration = {
    type = "fluentbit"

    options = {
      config-file-type        = "file"
[...]

PR #314, opened 2022-10-19
fix: Cannot specify more than one extra_container_definitions

dynamike:
## Description
Resolves issues defined in https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/269 when trying to pass in multiple container definitions into `extra_container_definitions`. The main issue is Terraform's requirement for exact matching objects when untyping a variable as `any` More detail is defined in https://github.com/hashicorp/terraform/issues/26265. For the time being, this PR untypes the variable to get around the error. A more robust solution would leverage required and optional attributes to define a container definition. 

## Motivation and Context
Resolves https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/269, which seems like an increasingly common issue with more and more sidecar containers used for things like log routing and observability.

## Breaking Changes
There shouldn't be any breaking changes. 

## How Has This Been Tested?


versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 0.13.1"
+  required_version = ">= 1.3.0"
```
bryantbiggs, line 2: This is a breaking change and we do not want to push users to the bleeding edge as it will reduce adoption

bryantbiggs (changes requested):
I think we can provide a solution without the need for a breaking change and pushing users to use the latest version of Terraform

bryantbiggs (approved):

bryantbiggs:
Thank you @dynamike

antonbabenko:
This PR is included in version 3.22.1 :tada:

PR #315, opened 2022-10-22
feat: Use vpc cidr for efs sg unless overridden

nitrocode:
## Description

- Use vpc cidr for efs sg unless overridden

## Motivation and Context

- The module allows passing in a `cidr` to create a vpc and use that same `cidr` as `ingress_cidr_blocks` to the efs sg.
- If a `vpc_id` is passed in, the hcl will still use the `cidr` input to set the efs sg
- This change will continue to let `cidr` input as the efs sg ingress cidr blocks but if it is left empty (default), the efs sg ingress cidr blocks will default to the vpcs cidr, either the vpc passed in or the vpc created.

## Breaking Changes

N/A

## How Has This Been Tested?

## references
- Thanks to @bodgit for the original fix in https://github.com/bodgit/terraform-aws-atlantis/commit/54bd2d9a043e0e82937aa1ba724e8a08677d3acf
- Closes https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/252

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,6 +1,7 @@
 locals {
   # VPC - existing or new?
   vpc_id             = var.vpc_id == "" ? module.vpc.vpc_id : var.vpc_id
+  cidr               = var.cidr == "" ? data.aws_vpc.this.cidr : var.cidr
   private_subnet_ids = coalescelist(module.vpc.private_subnets, var.private_subnet_ids, [""])
   public_subnet_ids  = coalescelist(module.vpc.public_subnets, var.public_subnet_ids, [""])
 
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -383,7 +388,8 @@ module "efs_sg" {
   vpc_id      = local.vpc_id
   description = "Security group allowing access to the EFS storage"
 
-  ingress_cidr_blocks = [var.cidr]
+  ingress_cidr_blocks = local.efs_sg_ingress_cidr
```
dynamike, line 391: Looking at this whole security group makes me think that you don't need to allow ingress to the EFS volume from the entire VPC. Instead we can just rely on https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/master/main.tf#L387-L390, which already allows nfs-tcp from the Atlantis security group.

dynamike (commented):
Dropped a comment on just removing the cidr ingress rule.

dynamike:
Closing in favor of https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/318

PR #328, opened 2022-12-30
feat: Add profile and region variables

Ileriayo:
## Description

This change adds profile and region variables so that the user can specify which AWS account on their machine they would like to use. It also updates the `terraform.tfvars.sample` to include variables for using GitHub apps instead of Personal Access Tokens.

## Motivation and Context

A user may have more than one AWS account and would like to use either. This change allows them to specify which AWS profile to use, other than the `default` profile.
It also allows you to specify a region rather than hard-coding it. If no region is specified, the original hard-coded region has been made the default.

## Breaking Changes

No breaking changes

## How Has This Been Tested?

### DIFF
--- a/examples/github-complete/main.tf
+++ b/examples/github-complete/main.tf
@@ -1,5 +1,5 @@
 locals {
-  name   = "atlantis"
+  name   = "github-complete"
   region = var.region
 
   tags = {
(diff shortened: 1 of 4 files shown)


bryantbiggs:
thank you for the PR but this is not a valid change - this is an example and users are responsible for configuring the provider to suite their requirements

PR #339, opened 2023-03-02
feat: Add EFS performance settings.

lazzurs:
## Description
Adding EFS performance settings as the default EFS performance settings
might not provide enough performance to initial Atlantis installs that
have to pull large repos while the EFS volume is small as the default
mode provides performance relative to disk usage.

## Motivation and Context
In testing the default EFS mode provided initial performance so far away
from ephemeral storage that Atlantis became unusable.

## Breaking Changes
None

## How Has This Been Tested?
The current examples do not use EFS so did not updated them with these 
parameters.


variables.tf
```diff
@@ -743,6 +743,22 @@ variable "efs_file_system_token" {
...
+  validation {
+    condition     = contains(["bursting", "provisioned", "elastic"], var.efs_throughput_mode)
+    error_message = "The mode must be one of \"bursting\" \"provisioned\" or \"elastic\""
+  }
+}
+
+variable "efs_provisioned_throughput_in_mibps" {
+  description = "The throughput, measured in MiB/s, that you want to provision for the file system. Only applicable with efs_throughput_mode set to provisioned"
+  type        = number
+  default     = null
```
Moncky, line 759: What happens if this is null and set to provisioned.  Is it worth adding validation to ensure that it is a valid number if efs_througpuhs_mode is set to provisioned?

Moncky:
In addition to my inline comment there is an important not regaring the use of the provisioned throuhgput type in the AWS docs https://docs.aws.amazon.com/efs/latest/ug/managing-throughput.html

It is probably prudent to add that note to our README as well

variables.tf
lazzurs: You cannot do validation on another variable within a single variable. I did try adding validation for exactly this and terraform validate failed with an error indicating this wasn't possible.

variables.tf
```diff
@@ -743,6 +743,22 @@ variable "efs_file_system_token" {
   default     = ""
 }
 
+variable "efs_throughput_mode" {
+  description = "(Optional) Throughput mode for the file system. Defaults to bursting. Valid values: bursting, provisioned, or elastic. When using provisioned, also set provisioned_throughput_in_mibps."
+  type        = string
+  default     = "bursting"
+  validation {
```
bryantbiggs, line 750: Please remove the validation

variables.tf
```diff
@@ -743,6 +743,22 @@ variable "efs_file_system_token" {
   default     = ""
 }
 
+variable "efs_throughput_mode" {
+  description = "(Optional) Throughput mode for the file system. Defaults to bursting. Valid values: bursting, provisioned, or elastic. When using provisioned, also set provisioned_throughput_in_mibps."
+  type        = string
+  default     = "bursting"
```
bryantbiggs, line 749: please default to `null`

bryantbiggs (changes requested):

variables.tf
lazzurs: Why remove the validation? There is another validation in this file.

bryantbiggs: its doubly redundant - the provider already checks that only select values are required and as new modes are added, they will be added at the provider without any code changes required here

lazzurs: Done, thanks for the response.

lazzurs:
I don't think the failing checks are the fault of this PR. Is there anything more I can do?

lordz-md:
This is definitely beneficial

antonbabenko:
Please fix the failing CI checks and it looks good to be merged.

lazzurs:
As far as I can see the CI checks that are failing have nothing to do with the changes i've made and are checking for older Terraform versions. Have I missed something?

antonbabenko:
@lazzurs https://github.com/terraform-aws-modules/terraform-aws-atlantis/actions/runs/4620640384/jobs/8171227679?pr=339 - the code in `./examples/gitlab-repository-webhook` is failing because of optimistic versions of the Gitlab provider. Please update `versions.tf` in that directory.

lordz-md:
> @lazzurs https://github.com/terraform-aws-modules/terraform-aws-atlantis/actions/runs/4620640384/jobs/8171227679?pr=339 - the code in `./examples/gitlab-repository-webhook` is failing because of optimistic versions of the Gitlab provider. Please update `versions.tf` in that directory.

@antonbabenko I think this is fixed in current master branch of module and @lazzurs just has to rebase from it and it should fix the issue.

antonbabenko:
This PR is included in version 3.27.0 :tada:

PR #343, opened 2023-05-04
Make the enable_nat_gateway and single_nat_gateway variables configurable

chtakahashi:
## Description
Our team recently received an email from AWS recommending the utilization of separate NAT Gateways per AZ, and we tracked down the VPC / NAT Gateway combinations to be coming from our usage of the atlantis module.
This PR adds variables to configure the arguments `enable_nat_gateway` and `single_nat_gateway` to the `terraform-aws-modules/vpc/aws` submodule.

## Motivation and Context
This change is in direct response to the following message we received from AWS, via e-mail:

> We have observed that your Amazon VPC resources are using a shared NAT Gateway across multiple Availability Zones (AZ). To ensure high availability and minimize inter-AZ data transfer costs, we recommend utilizing separate NAT Gateways in each AZ and routing traffic locally within the same AZ.
> 
> Each NAT Gateway operates within a designated AZ and is built with redundancy in that zone only. As a result, if the NAT Gateway or AZ experiences failure, resources utilizing that NAT Gateway in other AZ(s) also get impacted. Additionally, routing traffic from one AZ to a NAT Gateway in a different AZ incurs additional inter-AZ data transfer charges. We recommend choosing a maintenance window for architecture changes in your Amazon VPC.

## Breaking Changes
No changes. All new variables are optional, and I set the newly added variables to default to the same initial values as before.

## How Has This Been Tested?
We are using this in our own AWS environment.


chtakahashi:
Noted the pre-commit / min tf pre-commit is failing due to incompatible versions - this was unrelated to the changes I made. Looking for guidance on how to proceed here! 😃

bryantbiggs (approved):

antonbabenko:
This PR is included in version 3.28.0 :tada:

PR #351, opened 2023-08-08
feat: Add logic and vars to opt out of acm / https listener if desired

Nathan-Yorio:
## Description
Add three variables, bool true by default so as not to conflict with existing expected functionality:
create_acm_certificate
create_https_listener
atlantis_url_ssl

## Motivation and Context
There are scenarios where one may wish to have this be generated for testing and the requirement of an acm certificate and https listener is not necessary. For instance, security enforced by IP ingress restriction in a test/nonprod.

Of course, in a real deployment it makes sense to force that functionality
(hence leaving these things to bool true by default so that only a user who intends to not have ssl and knows what they're doing would toggle it to false using the variable)

---
When applying this, they'd need to add the following in order to get a no https LB listener

```
atlantis_url_ssl = false
create_acm_certificate=false
create_https_listener=false
```

## Breaking Changes
None that I can see.

## How Has This Been Tested?
- [ x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ x] I have tested and validated these changes using one or more of the provided `examples/*` projects
    - have tested this against github complete

- [ x] I have executed `pre-commit run -a` on my pull request
    - ran docker version
    - only issue was interpolation syntax, which I fixed in a commit

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,28 @@
 repos:
-- repo: https://github.com/antonbabenko/pre-commit-terraform
-  rev: v1.81.0 # Get the latest from: https://github.com/antonbabenko/pre-commit-terraform/releases
-  hooks:
-    - id: terraform_fmt
-    - id: terraform_docs
+  - repo: https://github.com/antonbabenko/pre-commit-terraform
+    rev: v1.76.0
+    hooks:
+      - id: terraform_fmt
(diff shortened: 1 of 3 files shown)


Nathan-Yorio:
additionally, I guess they could all be one bool... just depends on how piecewise you'd want that functionality to be in the case of only setting some, but not all of these

PR #368, opened 2023-11-09
fix: Correct EFS mount and correct listener authentication when value is not provided

### LINKED ISSUE #367: EFS Volumes not getting mounted
## Description

The EFS volume, is not getting mounted by the atlantis container.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: 4.0.0

- Terraform version: 1.5.2
- 
## Reproduction Code [Required]

https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/master/examples/github-separate/main.tf#L30

## Expected behavior

I would expect to have /home/atlantis mounted on the EFS volume

## Actual behavior

EFS volume is created, but nothing is mounting it to the container

## Additional context

```
  volume = { for k, v in merge(
    {
      name = "efs"
      efs_volume_configuration = {
        file_system_id     = module.efs.id
        transit_encryption = "ENABLED"
        authorization_config = {
          access_point_id = module.efs.access_points["atlantis"].id
          iam             = "ENABLED"
        }
      }
    },
    lookup(var.service, "volume", {})
  ) : k => v if var.enable_efs }
```

Render in JSON as

```
    "volumes": [
        {
            "name": "efs_volume_configuration",
            "host": {}
        },
        {
            "name": "name",
            "host": {}
        }
    ],
```

AtlantisVolumes

bryantbiggs:
## Description
- Correct EFS mount 
- Replace the https listeners `authenticate_cognito` and `authenticate_oidc` with an overriding argument of `https_listener` where those values can be provided by users

## Motivation and Context
- Resolves #367 
- Currently, when you deploy the module and values are not provided to one or both of `authenticate_cognito` or `authenticate_oidc` of the HTTPs listener, Terraform throws an error that a default action was not provided. This is because the argument keys are provide, but not values, and [the underlying ALB module](https://github.com/terraform-aws-modules/terraform-aws-alb/blob/e3586dd0769e9aead190b980a041405f258fb1cd/main.tf#L80) is only checking to see if the key is present to determine if it should create the associated configuration. This must have been a late change on v9.0 that wasn't caught, because the module currently does not work as intended

## Breaking Changes
- Technically, yes due to the change to `authenticate_cognito` and `authenticate_oidc` but since I cannot deploy the module as its written today and this is on the back of a major release I believe a patch release is sufficient

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -65,6 +65,7 @@ module "alb" {
       http-https-redirect = {
         port     = 80
         protocol = "HTTP"
+
         redirect = {
           port        = "443"
           protocol    = "HTTPS"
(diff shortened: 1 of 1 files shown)


antonbabenko (approved):

antonbabenko:
This PR is included in version 4.0.1 :tada:

danielpops:
Nice, I came here to report the cognito/oidc bug. thanks for fixing it before I could even report it :)

PR #373, opened 2023-11-17
Add ACM module outputs to top-level atlantis module outputs

danielpops:
## Description
Add the ACM module outputs to the top-level atlantis module outputs

## Motivation and Context
To add additional https listeners, I want to reference the ACM certificate ARN

## Breaking Changes
None

## How Has This Been Tested?
Modified locally and used as
```
module "atlantis"  {
  [...]

  alb = {
    listeners = {
      https_unauth = {
        port            = 8443
        protocol        = "HTTPS"
        ssl_policy      = "ELBSecurityPolicy-TLS13-1-2-Res-2021-06"
        certificate_arn = module.atlantis.acm.acm_certificate_arn
        forward = {
          target_group_key = "atlantis"
        }
      }
    }
   [...]
  }
  [...]
}
```

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -34,3 +34,12 @@ output "efs" {
   description = "EFS created and all of its associated outputs"
   value       = module.efs
 }
+
+################################################################################
+# ACM
+################################################################################
+
+output "acm" {
+  description = "ACM created and all of its associated outputs"
+  value       = module.acm
+}


bryantbiggs:
hmm, I think we'll need to think about this because this looks cyclical and I'm skeptical that it would work if applied from scratch. I'll have to see if we can have a self-referencing value since the cert is created internally

danielpops:
> hmm, I think we'll need to think about this because this looks cyclical and I'm skeptical that it would work if applied from scratch. I'll have to see if we can have a self-referencing value since the cert is created internally

Fair enough, I think I did not test it out from scratch, only after adding the additional listener. I'm also open to any other suggestions on how to achieve the goal of using the generated SSL certificate for additional listeners.

PR #375, opened 2023-11-30
fix: Correct ALB atlantis target group protocol and port arguments

pmorillas-apheris:
Following the upgrade to Terraform ALB module major version 9, it is essential to note that the target group arguments `backend_protocol` and `backend_port` now require adjustment. Going forward, these should be replaced with `protocol` and `port` respectively.


antonbabenko:
This PR is included in version 4.0.5 :tada:

Issue #377, opened 2023-11-30
ATLANTIS_ATLANTIS_URL not correctly picked up from atlantis.fqdn


MrPiao:
## Description

I am initializing the terraform-aws-atlantis module with the DNS records for the ALB manually configured (the domain we want is managed outside of Route53). To ensure `local.atlantis_url` is correctly set within the module, I am trying to set `var.atlantis.fqdn` to the correct URL; however, when I run `terraform plan`, the env variable for `ATLANTIS_ATLANTIS_URL` is set to just `https://`.

I have worked around this issue by defining another env variable with `ATLANTIS_ATLANTIS_URL => url we want`, but this is not ideal as AWS squashes the two env variables together, resulting in `terraform plan` always indicating that there is a change to made despite there not being a code change.

Relevant LOC: https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/f983fd8fb5ada3be696aeb4c4e1329de845a31f9/main.tf#L4

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: 4.0.3

- Terraform version: 1.6.3

- Provider version(s): aws 5.28.0

## Reproduction Code [Required]
```
module "atlantis" {
  source  = "terraform-aws-modules/atlantis/aws"
  version = "4.0.3"

  ...

  # ECS Container Definition
  atlantis = {
    fqdn  = "foobar.com"
    image = "ghcr.io/runatlantis/atlantis:v0.26.0"
    ...
  }

  ...

  create_certificate = false
  create_route53_records = false
}
```

## Expected behavior

Env variable `ATLANTIS_ATLANTIS_URL` should be set to "https://foobar.com"

### Plan output (only relevant parts included)
```
      ~ container_definitions    = jsonencode(
          ~ [
              ~ {
                  ~ environment            = [
                      + {
                          + name  = "ATLANTIS_ATLANTIS_URL"
                          + value = "https://"
                        },
                        # (8 unchanged elements hidden)
                    ]
```

remiflament:
Hello,
Did you test this issue with the last version of the module? `4.0.5`

This local env should do the job
```terraform
atlantis_url = "https://${try(coalesce(
    try(var.atlantis.fqdn, null),
    module.alb.route53_records["A"].fqdn,
    module.alb.dns_name,
  ), "")}"
```

[...]

PR #379, opened 2023-12-07
fix: FQDN detection when using an existing ALB

legoscia:
## Description
In the definition of the local variable `atlantis_url`, currently attributes of `module.alb` are referenced within a `coalesce` function call. `coalesce` isn't short-circuiting, so even if `var.atlantis.fqdn` was provided, it won't be used in the case when an existing ALB is used: the reference to `module.alb.route53_records["A"]` will fail, and the error will be caught by the top-level `try`, leaving `atlantis_url` as `https://`.

In this change, the individual references to `module.alb` are wrapped in `try` calls, meaning that `var.atlantis.fqdn` will be respected if provided.

## Motivation and Context
This change fixes this issue: https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/377
With this change, links in the "Checks" section of a Github pull request point to the correct URL.

## Breaking Changes
This change can force replacement of the ECS task definition in the case where an existing ALB was used and `var.atlantis.fqdn` was specified (but ignored). If this is not desirable, this can be avoided by removing that variable from the configuration. I believe this is unlikely to happen.

## How Has This Been Tested?
- [X] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [X] I have tested and validated these changes using one or more of the provided `examples/*` projects

  I added the `fqdn` variable to `github-separate`, and deployed it without this change. That resulted in `ATLANTIS_ATLANTIS_URL` being set to `https://`. Then I ran `terraform apply` with this change. That resulted in `module.atlantis.module.ecs_service.aws_ecs_task_definition.this[0]` being recreated and `ATLANTIS_ATLANTIS_URL` being set to the URL of the load balancer.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,8 +1,7 @@
 locals {
   # Atlantis
   atlantis_url = "https://${try(coalesce(
-    try(var.atlantis.fqdn, null),
-    module.alb.route53_records["A"].fqdn,
+    try(var.atlantis.fqdn, module.alb.route53_records["A"].fqdn, null),
     module.alb.dns_name,
   ), "")}"
 
--- a/examples/github-separate/main.tf
+++ b/examples/github-separate/main.tf
@@ -67,6 +67,7 @@ module "atlantis" {
         valueFrom = try(module.secrets_manager["github-webhook-secret"].secret_arn, "")
       },
     ]
+    fqdn = module.alb.dns_name
   }
 
   service = {


main.tf
```diff
@@ -2,8 +2,8 @@ locals {
   # Atlantis
   atlantis_url = "https://${try(coalesce(
```
bryantbiggs, line 3: I think this all can be re-written and simplified to:

```hcl
atlantis_url = "https://${try(
    var.atlantis.fqdn,
    module.alb.route53_records["A"].fqdn,
    module.alb.dns_name,
    "",
  )}"
```

legoscia: I tried that, but I got this error:
```
╷
│ Error: Invalid template interpolation value
│ 
│   on ../../main.tf line 3, in locals:
│    3:   atlantis_url = "https://${try(
│    4:     var.atlantis.fqdn,
│    5:     module.alb.route53_records["A"].fqdn,
│    6:     module.alb.dns_name,
│    7:     ""
│    8:   )}"
│     ├────────────────
│     │ module.alb.dns_name is null
│     │ module.alb.route53_records is object with no attributes
│     │ var.atlantis is object with no attributes
│ 
│ The expression result is null. Cannot include a null value in a string template.
╵
```

I think the problem is that `module.alb.dns_name` is always defined, even if it is `null`. Therefore, it doesn't throw an error, and `try` returns `null`.

I tried wrapping the `try` call in a call to `coalesce`, but that didn't fix the error, since `coalesce` raises an error when the final result is an empty string.

So I think we need the call to `coalesce` to drop the load balancer DNS name if it is null, and we need the `try` outside the `coalesce` to handle the case when there is no valid name. I removed the `try` around `module.alb.dns_name`, as it doesn't seem to be needed.

main.tf
```diff
@@ -2,7 +2,7 @@ locals {
...
     try(var.atlantis.fqdn, null),
```
bryantbiggs, line 4: hmm, ok. well we can at least save one `try()` for now with the following. I believe this should work if what you are proposing works:
```hcl
atlantis_url = "https://${try(coalesce(
    try(var.atlantis.fqdn, module.alb.route53_records["A"].fqdn, null),
    module.alb.dns_name,
  ), "")}"
```

legoscia: That's a good point. Just updated the commit - seems to work!

bryantbiggs (approved):
Thank you!

antonbabenko:
This PR is included in version 4.0.6 :tada:

PR #387, opened 2023-12-21
fix: Add 'Name' Tag to EFS File System

### LINKED ISSUE #386: When EFS is Enabled, the Created EFS File System has an empty 'Name' Tag
## Description

Specifying `enable_efs = true` causes the module to create an EFS file system, but it has an empty `Name` tag set. This is used by the AWS EFS management console uses for identifying the file system, so would be useful if it was set.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version: `4.0.6`
- Terraform version: `v1.5.5`

## Expected behavior

Set the EFS file system `Name` tag to the value of `var.name`.

## Actual behavior

Empty `Name` tag

### Terminal Output Screenshot(s)

image

X-Guardian:
## Description

Add `name = var.name` property to the EFS module. This value is added as a tag to the EFS file system by the EFS module.

## Motivation and Context

- Fixes #386 

## How Has This Been Tested?
- [x] I have tested this on my own configuration

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -466,6 +466,7 @@ module "efs" {
   version = "1.3.1"
 
   create = var.create && var.enable_efs
+  name   = var.name
 
   # File system
   availability_zone_name          = try(var.efs.availability_zone_name, null)


bryantbiggs (approved):
makes sense, thanks for catching!

antonbabenko:
This PR is included in version 4.0.7 :tada:

PR #391, opened 2024-01-03
feat: Configurable default action https listener

aalvarezaph:
Make the ALB https listener default action configurable

## Description

This PR adds a new variable `alb_https_default_action` that allows configuring the default action for the ALB https listener.

## Motivation and Context

In my scenario, I wanted the default action to be a blocking one (send a fixed response instead of forwarding to Atlantis target group).

The reason for it is that I'm adding a rule that grants access through OIDC Authentication, but for that to make sense I need to have a default rule that doesn't forward.

Currently, this default action is hardcoded so this PR adds a variable to allow overriding it.

## Breaking Changes

No breaking change, the new variable defaults to the value that preserves current behaviour

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ No resources.
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | <a name="input_alb"></a> [alb](#input\_alb) | Map of values passed to ALB module definition. See the [ALB module](https://github.com/terraform-aws-modules/terraform-aws-alb) for full list of arguments supported | `any` | `{}` | no |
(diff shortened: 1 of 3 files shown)


aalvarezaph:
@bryantbiggs sorry for tagging you, but is there anything else needed from my side?

bryantbiggs (approved):

antonbabenko:
This PR is included in version 4.1.0 :tada:

aalvarezaph:
@antonbabenko how is this propagated to Terraform registry? (registry still shows `4.0.8` as latest

bryantbiggs:
its usually automatic based on the tags created - looks like the registry didn't catch those or pick those up. Anton should be able to re-sync it when he gets a chance, looks like 4.2.0 should be latest

antonbabenko:
It is fixed by resyncing.

PR #392, opened 2024-01-05
feat: Add support for different name for EFS 

Andrews-repo:
## Description
The logic to name EFS was added, but it is tied to the main "name" variable for the module, and does not contain the logic to allow for uniques names when passing them in differently. 

## Motivation and Context
Declaring a unique name in the EFS array did not allow for setting unique names. Default/unique name for the module was applied from the Atlantis module.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -467,7 +467,7 @@ module "efs" {
   version = "1.3.1"
 
   create = var.create && var.enable_efs
-  name   = var.name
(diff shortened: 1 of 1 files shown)


bryantbiggs (approved):

antonbabenko:
This PR is included in version 4.2.0 :tada:

PR #409, opened 2024-12-20
feat: make load balancer optional

bmbferreira:
## Description

This makes the usage of a load balancer completely optional since it might not be needed to access atlantis. 

## Motivation and Context

Currently we are forcing the usage of a load balancer associated to the ecs service. That might not be needed since the access could be provided with something like cloudflare tunnels or AWS SSM sessions to make atlantis accessible from private networks without the need of a load balancer.

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,10 +235,10 @@ No resources.
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | <a name="input_alb"></a> [alb](#input\_alb) | Map of values passed to ALB module definition. See the [ALB module](https://github.com/terraform-aws-modules/terraform-aws-alb) for full list of arguments supported | `any` | `{}` | no |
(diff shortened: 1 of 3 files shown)


bryantbiggs:
we already support using an external load balancer - lets start with an issue and reproduction before jumping into a PR

bmbferreira:
> we already support using an external load balancer - lets start with an issue and reproduction before jumping into a PR

hey @bryantbiggs. Thanks but the what I'm trying to implement is to actually skip the load balancer entirely. As I explain on the `Motivation and Context` the module forces me to have a load balancer (either external or created within the module) and I don't want to have a load balancer at all, that's why I'm setting the `alb_target_group_arn` and `alb_security_group_id` as 100% optional to be able to not have a load balancer. I can put that information on an issue first if you prefer and associate with this PR. I hope I'm making myself clear but please let me know and I can add more details.

bmbferreira:
either way, created an issue here as recommended @bryantbiggs: https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/410

again, hope it's clear enough. Thank you.

PR #412, opened 2025-01-31
feat: Enable to set additional mount points

snovikov:
## Description

This will allow to add mount points in the container alongside EFS.

## Motivation and Context

When `enable_efs` is set to `true` it is not possible to set additional mount points for the container.
In case when you need to set `readonlyRootFilesystem` this could be necessary for accesing the container via session manager.

Links:
- https://toris.io/2021/06/using-ecs-exec-with-readonlyrootfilesystem-enabled-containers/
- https://github.com/aws/containers-roadmap/issues/1359

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -186,11 +186,11 @@ module "acm" {
 
 locals {
   mount_path = "/home/atlantis"
-  mount_points = var.enable_efs ? [{
+  mount_points = var.enable_efs ? concat(try(var.atlantis.mount_points, []), [{
     containerPath = local.mount_path
     sourceVolume  = "efs"
     readOnly      = false
-  }] : try(var.atlantis.mount_points, [])
+  }]) : try(var.atlantis.mount_points, [])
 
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -186,11 +186,11 @@ module "acm" {
 
 locals {
   mount_path = "/home/atlantis"
-  mount_points = var.enable_efs ? [{
+  mount_points = var.enable_efs ? concat(try(var.atlantis.mount_points, []), [{
```
bryantbiggs, line 189: I think this is such an edge case that if you want to do it, you might as well just provide all of the mount points (and therefore, no change required here)

snovikov: In case if  `enable_efs` is `true` - EFS volume and acces points are manged in the atlantis module, but `var.atlantis.mount_points` is ignored - it is not possible to provide all mount points as you suggested.

The only option would be to set `enable_efs` to `false` and manage EFS elsewhere.

PR #427, opened 2025-11-11
fix: Add missing `fqdn` attribute from the `atlantis` variable

albertorm95:
## Description

- Add missing `fqdn` attribute for variable `atlantis`

## Motivation and Context

- `fqdn` was missing in the `atlantis` variable so it was not being picked up by the:
  - ```hcl
    atlantis_url = "https://${try(coalesce(
      try(var.atlantis.fqdn, module.alb.route53_records["A"].fqdn, null),
      module.alb.dns_name,
    ), "")}"

## How Has This Been Tested?

- [X] I have deploy our current atlantis with this branch version and its working as expected


bryantbiggs:
> > Atlantis does not support running multiple copies at the same time, nor doe EFS allow setting values other than whats now hardcoded for the `deployment_*_percent` so I don't see us changing these values from whats hardcoded (unless someone can show a very valid reason to do so)
> > Why do you need exec access to the Atlantis container - this is a really large security concern given how much elevated access Atlantis typically has. Again, I don't see moving off the current value of having it disabled - that just seems like a very poor security decision
> 
> ok, I'll sync back with the team, I'll update the PR to just fix the `fqdn`

Thank you! Its not a hard no but more of a firm no unless theres a really good reason to change this. trying to reduce the number of choices/complexity to something that "just works" out of the box for users and does the "right" things

bryantbiggs (approved):

PR #428, opened 2025-11-11
fix: Update lookup logic for Atlantis URL due to variable optional attributes

bryantbiggs:
## Description
- Update lookup logic for Atlantis URL due to variable optional attributes

## Motivation and Context
- The current logic will resolve `try(var.atlantis.fqdn, module.alb.route53_records["A"].fqdn, null),` as `null` when not providing a `fqdn` because of the variable optional attributes (does not fall back to the next value on error, no error provided - just `null`). This coupled with the overall `coalesce()` means that the ALB DNS name is commonly exported instead of the URL using the Route53 record created by the module

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,9 +1,12 @@
 locals {
   # Atlantis
-  atlantis_url = "https://${try(coalesce(
-    try(var.atlantis.fqdn, module.alb.route53_records["A"].fqdn, null),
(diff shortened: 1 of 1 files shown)


antonbabenko:
This PR is included in version 5.0.2 :tada:

Issue #383, opened 2023-12-19
Redeploying fargate atlantis task always breaks


d33psky:
## Description

When fargate has an atlantis task running and a new deployment was issued the existing task prevents the new one from starting with this error : 

`Error: initializing server: starting BoltDB: timeout (a possible cause is another Atlantis instance already running
`

Note: EFS is being used.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Reproduction Code [Required]

Update terraform to redeploy the atlantis fargate task.

## Expected behavior

The new tasks starts and fargate stops the old task.

## Actual behavior

The new task fails with above BoltDB error which makes the new task fail, fargate retries starting a new task which also fails, etc.

### Terminal Output Screenshot(s)

`Error: initializing server: starting BoltDB: timeout (a possible cause is another Atlantis instance already running
`

## Workaround

After deploying an update to this module manually stop all running tasks and wait for fargate to start the new one.

bryantbiggs:
we can't control this behavior in the module - its up to users

mayn:
@bryantbiggs please consider reopening this as a bug or at minimum an opportunity to update docs for clarity. This occurs with the efs example from this module. It’s not really apparent until `wait_for_steady_state = true` at the service level or navigate to ecs task logs in AWS. 

Although atlantis itself and the use of boltdb emits the error, this module behavior is internally defaulting `deployment_minimum_healthy_percent = 66`.
1.  https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/3433ecf918056e3d347e774c81d3b878e51fbe18/main.tf#L236 

ECS will first launch new a task and the cutover. This works fine in when `enable_efs = false`, as the there is no shared filesystem. 

When `enable_efs = true` and an atlantis task beyond the initial deployment is provisioned , the deployment fails as described by the original author of this issue due to the shared filesystem + atlantis + boltdb

@d33psky , A work around, that seems to be working is to set `deployment_minimum_healthy_percent = 0` when `enable_efs = true`. This looks like it allows ECS to terminate the existing task, before launching the new deployment.
The cost is atlantis service is unavailable until the new deployment comes online. 

```
module “atlantis” {
    …
    # ECS service
    service = {
        deployment_minimum_healthy_percent = 0
    }
}

bryantbiggs:
ayyy, some proper information!

commarla:
[...]

PR #434, opened 2026-01-15
feat: Add variables for minimum/maximum deployment percentages

brandonstrohmeyer:
## Description

- Add `deployment_maximum_percent` field to `var.service`
- Add `deployment_minimum_healthy_percent` field to `var.service`
- Update `module.ecs_service` to use the new variables, replacing hardcoded attributes.

## Motivation and Context

Allowing users to set the deployment min/max percentage will allow for (near) zero downtime deployments provided the following prerequisites are met:
- EFS enabled
- Redis used in lieu of BoltDB

The prerequisites are a hard requirement, changing the deployment min/max without using Redis and EFS will result in an error (https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/383).

When `deployment_minimum_healthy_percent = 100` and `deployment_maximum_percent = 200` ECS will create the new Atlantis task before destroying the old container. 

A specific use case for this is allowing Atlantis to update its own configuration. When `var.wait_for_steady_state` is set,  the Terraform process will execute and apply all changes, waiting until the new task becomes healthy before exiting and terminating the container. In the event the new Atlantis task is not healthy, the `deployment_circuit_breaker` will trigger forcing a rollback and causing Terraform to exit with an error.

## Breaking Changes

No

## How Has This Been Tested?


albertorm95:
Nice, I try that, here:
- https://github.com/terraform-aws-modules/terraform-aws-atlantis/pull/427#issuecomment-3518400558

albertorm95:
@antonbabenko @bryantbiggs this is very useful, cloud we get a review?

bryantbiggs:
how is it useful? it was previously causing issues, and atlantis environments are ephemeral (though, you can save on EFS but not sure why thats valuable). seems like we're making things more complicated than they should be

brandonstrohmeyer:
> how is it useful? it was previously causing issues, and atlantis environments are ephemeral (though, you can save on EFS but not sure why thats valuable). seems like we're making things more complicated than they should be

I will expand on my use case in "Motivation and Context" above:

Our Atlantis deployment supports dozens if not hundreds of IaC repositories that deploy to many different environments. We have a lot of custom repo configuration logic that dictates which repos are permitted to deploy to specific environments. For more sensitive environments, we make use of allowlist patterns that dynamically build repo configuration rules at deploy time. Our entire repo config file uses HCL for construction and then is json encoded and passed to `ATLANTIS_REPO_CONFIG_JSON` at build time. 

Every change to the repo configuration requires a new deployment of Atlantis. 

Prior to the work above every time a change was requested, it required manual intervention by an engineer with appropriate permissions to checkout the merged Atlantis IaC and deploy. 

With the work above, we can streamline that process and support fully automated Atlantis configuration changes with a graceful rollback. User submits MR adding their required repo pattern to the appropriate config allowlist -> engineering team reviews and approves the MR -> User initiates `atlantis apply` and the change is automatically applied.

While it does add additional complexity, Atlantis is used in complex environments that require more than the average user may need. We have been using a forked version for several weeks without issue. 

It would be possible to make this feature more intuitive and user friendly by, instead of exposing these values as variables directly, include a "enable_atlantis_self_managed_mode" flag which handles creating all of the prerequisite infrastructure (i.e. Redis, EFS) and setting the correct environment parameters. I thought this was starting to creep out of scope, however.

PR #439, opened 2026-06-02
feat: Expose enable_execute_command in service variable

### LINKED ISSUE #436: Re-add ability to enable execute command pt. 2
## Is your request related to a new offering from AWS?

No. This is a request to re-expose the existing `enable_execute_command` parameter from the underlying `terraform-aws-ecs` module as an opt-in variable. The capability already exists in the ECS module ([`v6.7.0` — `variables.tf#L129-L134`](https://github.com/terraform-aws-modules/terraform-aws-ecs/blob/v6.7.0/modules/service/variables.tf#L129-L134)); it is currently hardcoded to `false` in this module ([`main.tf#L211`](https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/master/main.tf#L211)).

When upgrading to v5.0.0 (required for Terraform AWS Provider 6.x), the ability to use ECS Exec is permanently removed because the setting is hardcoded. There is no way to opt back in without forking the module.

Atlantis manages long-running Terraform operations, and real-world usage inevitably produces situations where container-level access is the only viable remediation path. Some examples gathered from our team and from [the community in #429](https://github.com/terraform-aws-modules/terraform-aws-atlantis/issues/429):

| Scenario | Why container access is needed |
|---|---|
| **Hung Terraform state** | Atlantis occasionally leaves Terraform processes in an inconsistent state. Killing the stuck process inside the container is the only way to unblock the stack without risking state corruption. |
| **Orphaned files on EFS** | When pull requests close unexpectedly, Atlantis may fail to clean up cloned repositories on the EFS volume. Over time this fills the volume; manual cleanup from within the container is the most reliable fix. |
| **Rotated credentials mid-operation** | If a Git token rotates after a pull request has been cloned, the old token is baked into the remote origin URL. The affected repos must be patched inside the running container to recover the hung PRs. |
| **Debugging startup or runtime failures** | Logs do not always capture the full picture — inspecting the filesystem, environment, or running processes inside the container is sometimes the fastest path to root cause. |

## We understand the security concern

The previous discussion in #429 raised a valid point: ECS Exec grants shell access to a role that often carries broad permissions, and enabling it by default would widen the attack surface.

**We fully agree.** That is exactly why we are asking for an **opt-in variable that defaults to `false`** — the secure-by-default posture is preserved, and teams that need temporary or controlled access can enable it deliberately, on their own terms.

This is consistent with how other sensitive settings are handled across the `terraform-aws-modules` ecosystem (e.g., `enable_nat_gateway`, `deletion_protection`, `publicly_accessible`) — the module exposes the knob; the deploying team owns the risk decision.

## Describe the solution you'd like

A minimal, backwards-compatible change:

**`variables.tf`** — add to the `service` variable object:

```hcl
enable_execute_command = optional(bool, false)
```

**`main.tf`** — drive the child module input from the variable:

```hcl
enable_execute_command = var.service.enable_execute_command
```

Default behaviour is unchanged. No existing deployments are affected.

## Describe alternatives we've considered

None of alternatives are sustainable for teams that rely on this module in production.

## Additional context

- Previous discussion: #429 (now locked)
- Multiple community members expressed the same need — this is not an edge case.

Thank you for maintaining this module — it's been a huge time-saver for our team, and we really appreciate the work that goes into it. We hope this makes a compelling case for reconsidering. 🙏

elkh510:
## Summary

Exposes the `enable_execute_command` parameter from the underlying `terraform-aws-ecs` service module as an optional, opt-in variable.

Previously this parameter was hardcoded to `false` in `main.tf`, making it impossible to enable ECS Exec without forking the module.

Changes:
- Add `enable_execute_command = optional(bool, false)` to the `service` variable object in `variables.tf`
- Pass `var.service.enable_execute_command` to the ECS service module in `main.tf`

Default value remains `false`, so this is fully backwards compatible.

Closes #436

## Use cases

- Kill hung Terraform processes without risking state corruption
- Clean up orphaned EFS files from failed repository clones
- Patch Git tokens mid-operation after credential rotation
- Debug failures by inspecting filesystem and processes when logs are insufficient

## Test plan

- [ ] `terraform validate` passes (verified locally)
- [ ] Default behavior unchanged (`enable_execute_command = false`)
- [ ] Opt-in works by setting `service = { enable_execute_command = true }`


elkh510:
Hey @antonbabenko @bryantbiggs, could you please take a look when you have a moment? Thank you!

bryantbiggs:
Thank you but as mentioned already ad nauseam, we are not going to re-add this security hole

